PR #5, opened 2026-04-15
[dev] Cache static Encoding in BlobString/BlobNullTerminatedString

### LINKED ISSUE #4: BlobString/BlobNullTerminatedString: new Encoding allocated per ToString() call
## Issue

`BlobString<TEncoding>.ToString()` and `BlobNullTerminatedString<TEncoding>.ToString()` allocate a new `TEncoding` instance on every call:

```csharp
// src/Paradise.BLOB/Blob/BlobString.cs
public new string ToString() => new TEncoding().GetString(Data.UnsafePtr, Data.Length);

// src/Paradise.BLOB/Blob/BlobNullTerminatedString.cs
public new string ToString() => new TEncoding().GetString(Data.UnsafePtr, Length);
```

`Encoding` instances are stateless and reusable. Creating a new one per call adds unnecessary allocation pressure, especially if `ToString()` is called frequently (e.g., in a game loop for debug display).

## Suggested fix

Cache a static instance:

```csharp
private static readonly TEncoding s_encoding = new TEncoding();
public new string ToString() => s_encoding.GetString(Data.UnsafePtr, Data.Length);
```

Note: Since `BlobString` and `BlobNullTerminatedString` are generic structs, `s_encoding` would be shared per `TEncoding` specialization, which is the desired behavior.

## Severity

**Low** — performance concern, no correctness impact. Each call allocates one `Encoding` object that becomes garbage immediately.

quabug:
Closes #4

## Summary
- Cache `TEncoding` instance as a `private static readonly` field (`s_encoding`) in both `BlobString<TEncoding>` and `BlobNullTerminatedString<TEncoding>`
- Eliminates unnecessary allocation of a new `Encoding` object on every `ToString()` call
- Since these are generic structs, `s_encoding` is shared per `TEncoding` specialization, which is the desired behavior

## Test plan
- [x] Paradise.BLOB builds with zero warnings (`TreatWarningsAsErrors=true`)
- [x] Paradise.BT builds with zero warnings (both `netstandard2.1` and `net10.0` targets)
- [x] All 484 BLOB tests pass

### DIFF
--- a/src/Paradise.BLOB/Blob/BlobNullTerminatedString.cs
+++ b/src/Paradise.BLOB/Blob/BlobNullTerminatedString.cs
@@ -4,10 +4,12 @@ namespace Paradise.BLOB;
 
 public unsafe struct BlobNullTerminatedString<TEncoding> where TEncoding : Encoding, new()
 {
+    private static readonly TEncoding s_encoding = new TEncoding();
+
     internal BlobArray<byte> Data;
     public int Length => Data.Length - 1/* length without termination */;
     public byte* UnsafePtr => Data.UnsafePtr;
-    public new string ToString() => new TEncoding().GetString(Data.UnsafePtr, Length);
+    public new string ToString() => s_encoding.GetString(Data.UnsafePtr, Length);
 #if UNITY_2021_2_OR_NEWER || NETSTANDARD2_1_OR_GREATER
     public System.Span<byte> ToSpan() => new System.Span<byte>(UnsafePtr, Length);
 #endif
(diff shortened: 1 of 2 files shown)


quabug:
## Review Summary
- **Iterations**: 1
- **Issues fixed**: none required
- **Remaining minor**: none

### Findings analysis
- Codex and Kimi-K2.5 raised thread-safety concerns about shared `Encoding` instances. This is a false positive: .NET `Encoding.GetString()` is documented as thread-safe, and `Encoding` instances are explicitly designed to be reusable across threads for decoding operations. Any custom `Encoding` that violates this contract would be breaking the base class's documented invariants.
- GLM-5 correctly identified the change as safe and approved.
- Qwen3.5-Plus flagged the `s_` naming prefix as non-standard, but `s_camelCase` is the project convention per `.editorconfig`.
- No critical or major issues identified.

(Gemini and MiniMax-M2.5 were unavailable due to API errors)

PR #6, opened 2026-04-15
[dev] Fix StructBuilder duplicate SetBuilder double-build

### LINKED ISSUE #3: StructBuilder: duplicate SetBuilder for same field causes double build
## Bug

`StructBuilder<T>.SetBuilder` maintains both a dictionary (`_fieldBuilderMap`) and a list (`_builders`). When called twice for the same field, the dictionary overwrites the old entry but the list appends without removing the old one:

```csharp
// src/Paradise.BLOB/Builder/StructBuilder.cs
public TBuilder SetBuilder<TField, TBuilder>(ref TField field, TBuilder builder)
    where TField : unmanaged
    where TBuilder : IBuilder<TField>
{
    var fieldOffset = _value.GetFieldOffset(ref field);
    _fieldBuilderMap[fieldOffset] = builder;  // overwrites old entry
    _builders.Add((fieldOffset, builder));     // appends — old entry remains
    return builder;
}
```

During `BuildImpl`, the `_builders` list is iterated, so both the old and new builder execute for the same field offset:

```csharp
protected override void BuildImpl(IBlobStream stream, ref T data)
{
    foreach (var (offset, builder) in _builders)
    {
        stream.Position = DataPosition + offset;
        builder.Build(stream);  // old builder writes first, then new builder overwrites
    }
}
```

While the final result may be correct (the new builder's output overwrites the old), the old builder still executes, which:
- Wastes work
- May have side effects (e.g., advancing `PatchPosition`)
- Can corrupt the blob if the old builder writes to the patch area

## Suggested fix

Either remove the old entry from `_builders` when the same field offset is set again, or replace the list with an ordered dictionary keyed by offset.

## Severity

**Medium** — correctness risk when builders have side effects on the stream's patch area.

quabug:
Closes #3

## Summary
- Fixed `StructBuilder<T>.SetBuilder` to remove the old entry from `_builders` when the same field offset is set again, preventing `BuildImpl` from executing both old and new builders
- Added tests verifying that only the latest builder's result appears in the final blob when `SetBuilder`/`SetArray`/`SetValue` is called twice for the same field

## Test plan
- [x] Existing 486 tests pass
- [x] New test `should_use_latest_builder_when_set_twice_for_same_field` — sets array field twice, verifies only second builder's data
- [x] New test `should_use_latest_builder_when_set_ptr_twice_for_same_field` — sets value field twice, verifies only second value

### DIFF
--- a/src/Paradise.BLOB.Test/TestBlobBuilder.cs
+++ b/src/Paradise.BLOB.Test/TestBlobBuilder.cs
@@ -194,6 +194,49 @@ public class TestBlobBuilder
         Assert.AreEqual(unicodeString, blob.Value.UnicodeStringPtr.Value.ToString());
     }
 
+    struct BlobWithArray
+    {
+        public int Value;
+        public BlobArray<int> Array;
+    }
+
+    [Test]
+    public void should_use_latest_builder_when_set_twice_for_same_field()
+    {
+        var builder = new StructBuilder<BlobWithArray>();
+        builder.SetValue(ref builder.Value.Value, 42);
+
+        // Set array field first time with [1, 2, 3]
+        builder.SetArray(ref builder.Value.Array, new int[] { 1, 2, 3 });
+
+        // Set array field second time with [10, 20, 30, 40, 50]
+        builder.SetArray(ref builder.Value.Array, new int[] { 10, 20, 30, 40, 50 });
+
+        var blob = builder.CreateManagedBlobAssetReference();
+
+        // Only the second builder's result should be present
+        Assert.AreEqual(42, blob.Value.Value);
+        Assert.AreEqual(5, blob.Value.Array.Length);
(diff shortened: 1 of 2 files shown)


quabug:
## Review Summary
- **Iterations**: 2
- **Issues fixed**: Renamed misleading test name (`set_ptr` -> `set_value`), removed redundant `SetValue(100)` call
- **Remaining minor**: `Is.EquivalentTo` used for array assertion (matches existing project pattern; values are distinct so order is implicitly tested); O(n) `RemoveAll` performance is acceptable for typical struct field counts

Reviewed by: Claude Code + Codex + Gemini (unavailable) + Qwen (GLM-5, Kimi-K2.5, MiniMax-M2.5 (unavailable), Qwen3.5-Plus)

PR #7, opened 2026-04-15
[dev] Fix dangling pointer in non-generic ManagedBlobAssetReference

### LINKED ISSUE #1: ManagedBlobAssetReference (non-generic): GetUnsafePtr returns dangling pointer
## Bug

The non-generic `ManagedBlobAssetReference` class returns a pointer from `GetUnsafePtr<T>()` that is only valid inside the `fixed` block, but the pointer escapes the block scope:

```csharp
// src/Paradise.BLOB/Blob/ManagedBlobAssetReference.cs
public T* GetUnsafePtr<T>() where T : unmanaged
{
    if (_blob.Length < sizeof(T)) throw new ArgumentException("invalid generic parameter");
    fixed (void* ptr = &_blob[0]) return (T*)ptr; // pointer is dangling after return
}
```

The `fixed` statement pins the `_blob` array only for the duration of the method. Once `GetUnsafePtr` returns, the GC is free to move the array, making the returned pointer dangling. Any caller that stores or dereferences this pointer after return gets undefined behavior.

The same issue affects `GetValue<T>()` which calls `GetUnsafePtr<T>()` and dereferences outside the fixed block:

```csharp
public ref T GetValue<T>() where T : unmanaged => ref *GetUnsafePtr<T>();
```

## Contrast with generic version

The generic `ManagedBlobAssetReference<T>` correctly pins the array for its entire lifetime using `GCHandle.Alloc(_blob, GCHandleType.Pinned)`. The non-generic version lacks this.

## Suggested fix

Either:
- Pin the byte array with `GCHandle.Alloc(Pinned)` like the generic version does, or
- Remove the non-generic class if its use cases are covered by the generic version

## Severity

**High** — any use of the returned pointer or ref is undefined behavior.

quabug:
Closes #1

## Summary
- Pin the byte array with `GCHandle` in the constructor so `GetUnsafePtr<T>()` returns a stable pointer
- Implement `IDisposable` with finalizer safety net and double-dispose protection
- Add tests for pointer stability, correct data access, double dispose, and empty-blob rejection

## Test plan
- [x] `GetUnsafePtr<T>()` returns a valid, stable pointer across multiple calls
- [x] `GetValue<T>()` returns correct data
- [x] `Dispose()` can be called twice without throwing
- [x] Constructor rejects empty blob array
- [x] `GetUnsafePtr<T>()` throws when blob is too small for type
- [x] Paradise.BLOB library builds with 0 warnings, 0 errors
- [x] No new errors introduced in test project (pre-existing test errors unchanged)

### DIFF
--- a/src/Paradise.BLOB.Test/TestManagedBlobAssetReference.cs
+++ b/src/Paradise.BLOB.Test/TestManagedBlobAssetReference.cs
@@ -1,0 +1,87 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace Paradise.BLOB.Test;
+
+public class TestManagedBlobAssetReference
+{
+    struct SimpleData
+    {
+        public int X;
+        public float Y;
+    }
+
+    [Test]
+    public unsafe void should_return_valid_pointer_from_GetUnsafePtr()
+    {
+        var data = new SimpleData { X = 42, Y = 3.14f };
+        var bytes = new byte[sizeof(SimpleData)];
+        fixed (byte* dst = bytes)
+        {
+            *(SimpleData*)dst = data;
+        }
+
+        using var blob = new ManagedBlobAssetReference(bytes);
+        var ptr = blob.GetUnsafePtr<SimpleData>();
+
+        Assert.AreEqual(42, ptr->X);
+        Assert.AreEqual(3.14f, ptr->Y);
+    }
+
+    [Test]
+    public unsafe void should_return_stable_pointer_across_calls()
+    {
+        var bytes = new byte[sizeof(int)];
+        fixed (byte* dst = bytes)
+        {
+            *(int*)dst = 123;
+        }
+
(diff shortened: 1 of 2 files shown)


quabug:
## Review Summary
- **Iterations**: 2 (1 fix round)
- **Issues fixed**: Added `_handle.IsAllocated` guard in finalizer to prevent crash if GCHandle was never allocated
- **Remaining minor**: 
  - Use-after-dispose check not added to `GetUnsafePtr<T>()` (consistent with generic `ManagedBlobAssetReference<T>`)
  - Thread-safety not added to disposal (standard for non-thread-safe disposable types)
  - XML docs and `sealed` modifier omitted (not in spec)

Reviewed by: Claude Code + Qwen (GLM-5, Kimi-K2.5, Qwen3.5-Plus)
Note: Codex and Gemini were unavailable due to API issues during review.

PR #8, opened 2026-04-15
[dev] Add finalizer and double-dispose protection to generic ManagedBlobAssetReference

### LINKED ISSUE #2: ManagedBlobAssetReference<T>: missing finalizer and no double-dispose protection
## Bug

`ManagedBlobAssetReference<T>` pins a byte array via `GCHandle.Alloc(Pinned)` but has two resource management problems:

### 1. No finalizer — GCHandle leaks if Dispose is not called

```csharp
// src/Paradise.BLOB/Blob/ManagedBlobAssetReference.cs
public unsafe class ManagedBlobAssetReference<T> : IDisposable where T : unmanaged
{
    private GCHandle _handle;

    public ManagedBlobAssetReference(byte[] blob)
    {
        _handle = GCHandle.Alloc(_blob, GCHandleType.Pinned);
    }

    public void Dispose() { _handle.Free(); }
    // No finalizer — if Dispose() is never called, the GCHandle leaks permanently
}
```

If a caller forgets to call `Dispose()` (or fails to use `using`), the `GCHandle` is never freed. Pinned GCHandles are particularly expensive because they prevent GC compaction of the pinned memory region.

### 2. No double-dispose protection

Calling `Dispose()` twice calls `_handle.Free()` twice, which throws `InvalidOperationException`. The standard dispose pattern guards against this:

```csharp
// Current code — throws on double dispose:
public void Dispose() { _handle.Free(); }
```

## Suggested fix

```csharp
private bool _disposed;

~ManagedBlobAssetReference() => Dispose();

public void Dispose()
{
    if (_disposed) return;
    _disposed = true;
    _handle.Free();
    GC.SuppressFinalize(this);
}
```

## Severity

**Medium** — resource leak if Dispose pattern is not followed; crash on double dispose.

quabug:
Closes #2

## Summary
- Added `_disposed` field and double-dispose guard to `ManagedBlobAssetReference<T>.Dispose()`
- Added `~ManagedBlobAssetReference()` finalizer to free `GCHandle` if `Dispose()` is never called
- Added `GC.SuppressFinalize(this)` in `Dispose()` to avoid redundant finalization
- Added tests for construction, value/pointer access, and double-dispose safety

## Test plan
- [x] `Dispose()` can be called multiple times without throwing
- [x] `UnsafePtr` and `Value` work correctly before dispose
- [x] Basic construction and value access still works
- [x] All 487 existing tests pass

### DIFF
--- a/src/Paradise.BLOB.Test/TestManagedBlobAssetReference.cs
+++ b/src/Paradise.BLOB.Test/TestManagedBlobAssetReference.cs
@@ -1,0 +1,51 @@
+using System.Runtime.InteropServices;
+
+namespace Paradise.BLOB.Test;
+
+public class TestManagedBlobAssetReference
+{
+    struct SimpleValue
+    {
+        public int X;
+        public float Y;
+    }
+
+    private static byte[] CreateBlob(int x, float y)
+    {
+        var builder = new ValueBuilder<SimpleValue>();
+        builder.Value.X = x;
+        builder.Value.Y = y;
+        return builder.CreateBlob();
+    }
+
+    [Test]
+    public unsafe void should_construct_and_access_value()
+    {
+        var blob = CreateBlob(42, 3.14f);
+        using var reference = new ManagedBlobAssetReference<SimpleValue>(blob);
+
+        Assert.AreEqual(42, reference.Value.X);
+        Assert.AreEqual(3.14f, reference.Value.Y);
+    }
+
+    [Test]
+    public unsafe void should_access_via_unsafe_ptr()
+    {
+        var blob = CreateBlob(99, 1.5f);
+        using var reference = new ManagedBlobAssetReference<SimpleValue>(blob);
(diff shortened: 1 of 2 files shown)


quabug:
## Review Summary
- **Iterations**: 2 (initial + fix for IsAllocated guard)
- **Issues fixed**: Added `_handle.IsAllocated` check before `_handle.Free()` to handle edge case where constructor fails partially
- **Remaining minor**: Thread-safety of `_disposed` field (low risk — `GCHandle.Free()` validates internally, and `IsAllocated` provides defense-in-depth); non-standard dispose pattern (YAGNI — class is not designed for inheritance)

Reviewed by: Claude Code + Qwen (GLM-5, Kimi-K2.5, Qwen3.5-Plus)
(Codex, Gemini, MiniMax-M2.5 unavailable due to API errors)

PR #13, opened 2026-04-15
[dev] Fix TickChildrenReturnBreakOrDefault returning 0 for completed children

### LINKED ISSUE #11: TickChildrenReturnBreakOrDefault returns 0 instead of completed child's state
## Bug

`NodeExtensions.TickChildrenReturnBreakOrDefault` sets `currentState = 0` for already-completed children, then returns that 0 if the break check matches:

```csharp
// src/Paradise.BT/Nodes/NodeExtensions.cs
currentState = tickCheck(previousState) ? VirtualMachine.Tick(childIndex, ref blob, ref bb) : 0;
if (breakCheck(currentState == 0 ? previousState : currentState))
{
    break;
}
// ...
return currentState;  // Returns 0 when child was skipped
```

The `breakCheck` correctly uses `previousState` when `currentState == 0`, but the method still returns `currentState` (which is 0) instead of `previousState`.

## Impact

Affects both `SequenceNode` and `SelectorNode`. When re-ticked after completion (with `AutoResetOnCompletion = false`), they return `(NodeState)0` instead of the actual result, breaking parent node logic.

Example: A Sequence where child A already failed → `breakCheck(Failure)` is true → breaks → returns `currentState = 0` instead of `Failure`.

## Suggested fix

Return the actual state that triggered the break:

```csharp
var effectiveState = currentState == 0 ? previousState : currentState;
if (breakCheck(effectiveState))
{
    return effectiveState;
}
```

## Evidence

Found by GLM-5 and Kimi-K2.5, verified by manual code review. Same root cause as the ParallelNode issue (#5) — composite nodes losing completed children's states.

## Severity

**Medium** — only manifests when trees are ticked after completion without reset, which is uncommon with the default `AutoResetOnCompletion = true`.

quabug:
Closes #11

## Summary
- Fix `TickChildrenReturnBreakOrDefault` to compute effective state (`currentState` or `previousState` for skipped children) once and use it for both the break check and return value
- Previously, when a child was already completed (skipped), `currentState` was set to `0` and that `0` was returned instead of the child's actual `previousState`
- Affects `SequenceNode` and `SelectorNode` when re-ticked after completion with `AutoResetOnCompletion = false`

## Test plan
- [x] Added `Sequence_Returns_Failure_Not_Zero_When_Child_Already_Failed` test
- [x] Added `Selector_Returns_Success_Not_Zero_When_Child_Already_Succeeded` test
- [x] All 14 tests pass
- [x] Build passes with warnings-as-errors

### DIFF
--- a/src/Paradise.BT.Test/BehaviorTreeTests.cs
+++ b/src/Paradise.BT.Test/BehaviorTreeTests.cs
@@ -147,6 +147,42 @@ public sealed class BehaviorTreeTests
     }
 
     [Test]
+    public async Task Sequence_Returns_Failure_Not_Zero_When_Child_Already_Failed()
+    {
+        var tree = BehaviorTreeBuilder.Build(
+            BehaviorNodes.Sequence(
+                BehaviorNodes.Failure(),
+                BehaviorNodes.Success()));
+
+        BehaviorTreeInstance instance = tree.CreateInstance(new Blackboard());
+        instance.AutoResetOnCompletion = false;
+
+        // First tick: Failure child breaks the sequence -> returns Failure
+        await Assert.That(instance.Tick()).IsEqualTo(NodeState.Failure);
+
+        // Second tick: child is already completed (Failure), should still return Failure not 0
+        await Assert.That(instance.Tick()).IsEqualTo(NodeState.Failure);
+    }
+
+    [Test]
+    public async Task Selector_Returns_Success_Not_Zero_When_Child_Already_Succeeded()
+    {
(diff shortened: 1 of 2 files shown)


quabug:
## Review Summary
- **Iterations**: 1
- **Issues fixed**: Added test for no-break completion path (all children already succeeded) per Codex feedback
- **Remaining minor**: Variable naming suggestions (lastState vs effectiveState) — kept `lastState` for consistency with `TickChildrenReturnLastOrDefault`. ParallelNode has same bug pattern — tracked separately in #5.

PR #15, opened 2026-04-15
[dev] Fix Blackboard by-value bug — return by ref

### LINKED ISSUE #12: BehaviorTreeInstance.Blackboard returns struct by value — mutations before first tick are lost
## Bug

`BehaviorTreeInstance.Blackboard` returns the `Blackboard` struct by value:

```csharp
// src/Paradise.BT/Core/BehaviorTreeInstance.cs:24
public Blackboard Blackboard => _blackboard;
```

`Blackboard` is a struct that lazily allocates its backing `BlackboardStorage` on first use. When a caller does:

```csharp
var instance = tree.CreateInstance();
instance.Blackboard.SetData(new MyData { Value = 42 }); // mutates a COPY — silently lost
instance.Tick(); // MyData is not in the blackboard
```

The `SetData` call creates a `BlackboardStorage` on the temporary copy returned by the property. The original `_blackboard` field still has `_storage = null`. The data is silently lost.

After the first `Tick()` call (which calls `_blackboard.SetData(new BehaviorTreeTickDeltaTime(...))` internally), `_storage` is initialized on the real field. Subsequent `instance.Blackboard` copies share the same `BlackboardStorage` reference, so mutations start working. This creates a confusing ordering dependency.

## Impact

Users seeding blackboard data before the first tick will silently lose their data. The behavior changes depending on whether `Tick()` was called first, making this a hard-to-debug correctness trap.

## Suggested fix

Either:
- Return `Blackboard` by `ref`: `public ref Blackboard Blackboard => ref _blackboard;`
- Or initialize `_storage` eagerly in the constructor

## Evidence

Found by Codex (GPT-5.4), verified by manual code review.

## Severity

**High** — silent data loss with no error, behavior depends on call order.

quabug:
Closes #12

## Summary
- Changed `BehaviorTreeInstance.Blackboard` property from by-value return to `ref` return
- This prevents silent data loss when calling `SetData()` before the first `Tick()`
- The `Blackboard` struct lazily allocates its backing `BlackboardStorage` — a by-value copy would allocate storage on a temporary that gets discarded

## Test plan
- Added `Blackboard_Mutations_Before_First_Tick_Are_Preserved` test that sets blackboard data before the first tick and verifies a custom node can read it
- All 13 existing + new tests pass
- Build passes with warnings-as-errors

### DIFF
--- a/src/Paradise.BT.Test/BehaviorTreeTests.cs
+++ b/src/Paradise.BT.Test/BehaviorTreeTests.cs
@@ -7,6 +7,23 @@ public sealed class BehaviorTreeTests
         public int Value;
     }
 
+    private struct PreTickData
+    {
+        public int Value;
+    }
+
+    [System.Runtime.InteropServices.Guid("F4E3D2C1-B0A9-4867-8765-432109FEDCBA")]
+    private struct ReadBlackboardNode : INodeData
+    {
+        public NodeState Tick<TNodeBlob, TBlackboard>(int index, ref TNodeBlob blob, ref TBlackboard bb)
+            where TNodeBlob : struct, INodeBlob
+            where TBlackboard : struct, IBlackboard
+        {
+            var data = bb.GetData<PreTickData>();
+            return data.Value == 42 ? NodeState.Success : NodeState.Failure;
+        }
+    }
+
     [System.Runtime.InteropServices.Guid("A1523157-2737-48A0-8F1D-14D07B5F4D77")]
     private struct CountingNode : INodeData
     {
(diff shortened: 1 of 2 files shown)


quabug:
## Review Summary
- **Iterations**: 1
- **Issues fixed**: None required — all findings are acknowledged design trade-offs
- **Remaining minor**: None

### Findings from multi-AI review:
1. **API breaking change** (raised by all) — Intentional per issue spec. Project is pre-1.0 with no external consumers.
2. **Exposed mutable state via ref** (Codex) — Standard C# ref return semantics. Callers using `var bb = instance.Blackboard` still get a copy. Direct `instance.Blackboard.SetData(...)` now correctly mutates the field.
3. **`var bb` aliasing still copies** (Codex) — Inherent to struct semantics in C#, not new or worsened by this change. The fix addresses the primary reported bug pattern.
4. **False claim that ref return prevents variable storage** (Qwen3.5-Plus) — Incorrect; C# allows storing ref returns in regular variables (creates a copy).

Reviewed by: Claude Code + Codex + Gemini (unavailable) + Qwen (GLM-5, Kimi-K2.5, MiniMax-M2.5 (unavailable), Qwen3.5-Plus)

PR #16, opened 2026-04-15
[dev] Fix ParallelNode returning invalid NodeState 0 for completed children

### LINKED ISSUE #9: ParallelNode returns invalid NodeState 0 when all children already completed
## Bug

`ParallelNode.Tick()` skips ticking completed children but ORs `0` into `flags` instead of the child's actual state:

```csharp
// src/Paradise.BT/Nodes/Composites/ParallelNode.cs:16
flags |= previousState.IsCompleted() ? 0 : VirtualMachine.Tick(childIndex, ref blob, ref bb);
```

When all children are already completed (e.g., on a subsequent tick without reset), `flags` remains `0` and the node returns `(NodeState)0` — which is not a valid state value (valid: Success=1, Failure=2, Running=4).

## Impact

- Parent nodes receiving `0` behave unpredictably
- `BehaviorTreeInstance.Tick()` checks `Status.IsCompleted()` for auto-reset, but `0` is NOT completed, so the tree gets stuck — never reset, never progressing
- Any multi-tick parallel scenario where children complete at different times loses the completed children's state contributions

## Suggested fix

Preserve completed children's states in the flags:

```csharp
flags |= previousState.IsCompleted() ? previousState : VirtualMachine.Tick(childIndex, ref blob, ref bb);
```

## Evidence

Found independently by 4/4 responding AI agents (GLM-5, Kimi-K2.5, MiniMax-M2.5, Qwen3.5-Plus) and verified by manual code review.

## Severity

**High** — incorrect behavior tree execution in any multi-tick parallel scenario.

quabug:
Closes #9

## Summary
- Fixed `ParallelNode.Tick()` OR-ing `0` into flags for already-completed children instead of preserving their actual state (`previousState`)
- Changed the empty-node fallthrough return from `0` to `NodeState.Success` to match convention that an empty composite succeeds
- Added two tests: `Parallel_Preserves_Completed_Children_State` (multi-tick scenario) and `Parallel_All_Children_Already_Completed_Returns_Valid_State` (re-tick without auto-reset)

## Test plan
- [x] `Parallel_Preserves_Completed_Children_State` — verifies completed child's Success state is preserved across ticks
- [x] `Parallel_All_Children_Already_Completed_Returns_Valid_State` — verifies re-tick with `AutoResetOnCompletion=false` returns valid Failure, not 0
- [x] Existing `Parallel_Returns_Failure_When_Any_Child_Fails_And_None_Are_Running` test still passes
- [x] All 14 BT tests pass

### DIFF
--- a/src/Paradise.BT.Test/BehaviorTreeTests.cs
+++ b/src/Paradise.BT.Test/BehaviorTreeTests.cs
@@ -174,6 +174,24 @@ public sealed class BehaviorTreeTests
     }
 
     [Test]
+    public async Task Parallel_Preserves_Failed_Child_State_Alongside_Running_Child()
+    {
+        // Child 1: instant Failure. Child 2: Running on tick 1, Success on tick 2.
+        var tree = BehaviorTreeBuilder.Build(
+            BehaviorNodes.Parallel(
+                BehaviorNodes.Failure(),
+                BehaviorNodes.Node(new CountingNode())));
+
+        BehaviorTreeInstance instance = tree.CreateInstance(new Blackboard());
+
+        // Tick 1: child1 Failure + child2 Running → Running (Running takes priority)
+        await Assert.That(instance.Tick()).IsEqualTo(NodeState.Running);
+
+        // Tick 2: child1 already completed (Failure preserved), child2 completes (Success) → Failure
+        await Assert.That(instance.Tick()).IsEqualTo(NodeState.Failure);
+    }
+
+    [Test]
     public async Task Custom_Struct_Node_Can_Be_Authored_Through_Interface_Constraints()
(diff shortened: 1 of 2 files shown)


quabug:
## Review Summary
- **Iterations**: 1
- **Issues fixed**: Added test for Failure+Running child combination (from Codex review)
- **Remaining minor**: 
  - Same invalid-state-on-completed-children pattern exists in Sequence/Selector via NodeExtensions (separate issue, out of scope for #9)
  - Empty Parallel fallback returns Success (dead code for non-empty parallel, acceptable edge case)

(Gemini and MiniMax-M2.5 were unavailable due to API errors)

PR #29, opened 2026-04-18
Make BehaviorTreeInstance generic over TBlackboard

quabug:
## Summary
- **`BehaviorTreeInstance<TBlackboard>`** — parameterised over `TBlackboard : struct, IBlackboard`. Field, ctor, `Blackboard` property, `Tick`, and `Reset` all pivot to `TBlackboard`. The `struct` constraint matches `VirtualMachine`'s generic signatures, so the tick pipeline stays allocation-free and the JIT specialises per concrete blackboard type.
- **`IBlackboard.SetData<T>(T)`** — added to the interface. Required because `Tick` writes `BehaviorTreeTickDeltaTime` through the generic blackboard. `Blackboard` already implements it; no new code there.
- **`BehaviorTree.CreateInstance`** — two overloads now:
  - `CreateInstance(Blackboard = default) → BehaviorTreeInstance<Blackboard>` preserves the existing default-Blackboard ergonomics (no call-site changes for callers passing the default blackboard).
  - `CreateInstance<TBlackboard>(TBlackboard) → BehaviorTreeInstance<TBlackboard>` for caller-supplied blackboard structs.
- **Sample + tests** — call sites that explicitly typed `BehaviorTreeInstance instance` now use `BehaviorTreeInstance<Blackboard> instance` (mechanical sed).

## Motivation
Lets projects layer their own blackboard struct (domain data, ECS components, etc.) on top of Paradise.BT without falling back to the boxed-interface path. Previously the engine was locked to the concrete `Blackboard` struct; now the concrete type is just the default.

## Test plan
- [x] `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` — 0 errors.
- [x] `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false --output normal` — all suites pass (BLOB, BT, Generators).
- [x] `dotnet run --project src/Paradise.BT.Sample/Paradise.BT.Sample.csproj` — sample behaves identically.

Generated with [Claude Code](https://claude.ai/code)


quabug:
### Response to iteration-1 review

**Fixed in acba099:**

- **[major] `IBlackboard.SetData<T>` widens the interface** — reverted. `IBlackboard` is back to its original read-only shape. Added a new **`IMutableBlackboard : IBlackboard`** carrying `SetData<T>(T)` for the cases that need to write (the tick path needs to push `BehaviorTreeTickDeltaTime`). `Blackboard` switches to `IMutableBlackboard` (it already had the method). External `IBlackboard` implementers are unaffected.
- **[major] Non-generic `BehaviorTreeInstance` API removed** — restored. `BehaviorTreeInstance<TBlackboard>` is now unsealed and a non-generic `BehaviorTreeInstance : BehaviorTreeInstance<Blackboard>` compat subclass keeps the original name + shape. `BehaviorTree.CreateInstance(Blackboard = default)` returns the non-generic form, so pre-refactor call sites compile unchanged. Reverted the Sample/Test sed back to `BehaviorTreeInstance instance` to exercise the compat path.
- **[minor] No test for the custom-`TBlackboard` path** — added `GenericBlackboardTests` with a `CountingBlackboard : IMutableBlackboard` struct. Two tests assert that `Tick` routes `SetData` through the user-supplied struct (counter increments), and that trees run to completion against a custom blackboard.

**Constraint update:** `CreateInstance<TBlackboard>` now constrains to `IMutableBlackboard` (tightened from `IBlackboard`) — the tick path writes, so the narrower constraint is actually correct.

quabug:
@opencara review

### Iteration 2 — refactor: delta-time propagation is now a caller concern

Pushed `ce3326b`.

- `BehaviorTreeInstance<TBlackboard>.Tick()` drops its `float` parameter and no longer writes `BehaviorTreeTickDeltaTime`. The instance core shouldn't know about one specific tick-input struct — that's a convention of time-based nodes, not an engine invariant.
- Class + `BehaviorTree.CreateInstance<T>` constraint relaxes from `IMutableBlackboard` back to `IBlackboard`. Read-only blackboard structs are now viable for the full runtime, not just observer use cases.
- Sample inlines the per-tick `SetData` before `Tick()` — the honest caller-side pattern.
- Tests pick up a test-local `TestTickExtensions.Tick(float)` extension that preserves the ergonomic syntax without shipping it as library API. It constrains to `IMutableBlackboard` since it writes.

Net effect: `IMutableBlackboard` disappears from the library hot path; it survives only as an API-taxonomy hint and the constraint on the test extension.

quabug:
@opencara review

PR #38, opened 2026-04-22
Port ParadiseECS PR #56 job system

quabug:
## Summary
- port the ParadiseECS job system from ParadiseEngine/ParadiseECS#56 into ParadiseEngine
- add Paradise.ECS.Jobs and Paradise.ECS.Jobs.Test with JobWorkerPool, WorkStealingPool, and wave schedulers
- wire IWorkItem, benchmarks, and solution/project references into this repo's ECS layout

## Source
- Original PR: https://github.com/ParadiseEngine/ParadiseECS/pull/56

## Verification
- dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true
- dotnet test --solution ParadiseEngine.slnx --no-build --output normal

### DIFF
--- a/ParadiseEngine.slnx
+++ b/ParadiseEngine.slnx
@@ -16,6 +16,7 @@
     <Project Path="src/Paradise.BT.Nodes/Paradise.BT.Nodes.csproj" />
     <Project Path="src/Paradise.BT/Paradise.BT.csproj" />
     <Project Path="src/Paradise.ECS.Concurrent/Paradise.ECS.Concurrent.csproj" />
+    <Project Path="src/Paradise.ECS.Jobs/Paradise.ECS.Jobs.csproj" />
     <Project Path="src/Paradise.ECS.Tag/Paradise.ECS.Tag.csproj" />
     <Project Path="src/Paradise.ECS/Paradise.ECS.csproj" />
   </Folder>
@@ -25,6 +26,7 @@
     <Project Path="src/Paradise.BT.Test/Paradise.BT.Test.csproj" />
     <Project Path="src/Paradise.ECS.Concurrent.Test/Paradise.ECS.Concurrent.Test.csproj" />
     <Project Path="src/Paradise.ECS.CoyoteTest/Paradise.ECS.CoyoteTest.csproj" />
+    <Project Path="src/Paradise.ECS.Jobs.Test/Paradise.ECS.Jobs.Test.csproj" />
     <Project Path="src/Paradise.ECS.Generators.Test/Paradise.ECS.Generators.Test.csproj" />
     <Project Path="src/Paradise.ECS.Test/Paradise.ECS.Test.csproj" />
   </Folder>
--- a/src/Paradise.ECS.Benchmarks/BenchmarkComponents.cs
+++ b/src/Paradise.ECS.Benchmarks/BenchmarkComponents.cs
@@ -1,0 +1,13 @@
+namespace Paradise.ECS.Benchmarks;
+
+[Component("EAEEB062-5E8F-4B6A-B3B6-1F60393B3547", Id = 0)]
+public partial struct BenchPosition
+{
+    public float X, Y, Z;
+}
+
+[Component("897422E5-00F8-47FE-BB9B-9035DD17E0A3", Id = 1)]
+public partial struct BenchVelocity
+{
+    public float X, Y, Z;
+}
--- a/src/Paradise.ECS.Benchmarks/BenchmarkSystems.cs
+++ b/src/Paradise.ECS.Benchmarks/BenchmarkSystems.cs
@@ -1,0 +1,20 @@
+namespace Paradise.ECS.Benchmarks;
+
+/// <summary>
+/// Benchmark system: adds velocity to position per entity.
+/// </summary>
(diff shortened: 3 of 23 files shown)


quabug:
@opencara review

quabug:
## Response to OpenCara review (2026-04-22)

Thanks for the thorough review. Findings #2 (SpinWait), #3 (`_workAvailable` reset race), and #4 (ARM64 ordering) are good catches and will be addressed in this push. Finding #5 will be answered with an inline csproj comment.

However, I have to **decline finding #1 (CRITICAL)** after analysis. The proposed fix would introduce a real data race; the original `Steal()` is canonically correct per Chase-Lev. Detailed reasoning below.

### Why the original `Steal()` is correct

Original code:
```csharp
int t = Volatile.Read(ref _top);
int b = _bottom;
if (b - t <= 0) return -1;

var buffer = _buffer;                       // capture BEFORE CAS
int item = buffer[t & (buffer.Length - 1)];

if (Interlocked.CompareExchange(ref _top, t + 1, t) != t) return -1;
return item;
```

The key invariant the reviewer's analysis missed: **`Grow()` never modifies the old buffer.** It allocates a new array, copies elements into it, then swaps `_buffer = newBuf`. Two cases:

- **Stealer captures `oldBuf`, then owner Grows:** `oldBuf[t & (oldLen-1)]` is unchanged forever (owner's subsequent pushes go to `newBuf`, not back to `oldBuf`). Read returns the correct value.
- **Owner Grows first, then stealer captures `newBuf`:** `Grow()` was called while `_top == t` (we hadn't CAS'd yet), so it copied slot `t` from `oldBuf` to `newBuf[t & (newLen-1)]`. Read returns the correct value.

In both cases the stealer reads the value that was logically at index `t` when it observed `b - t > 0`.

### Why the proposed fix is BROKEN

Concrete counter-example with proposed `var buffer = _buffer` after the CAS, initial capacity = 2, items pushed `A, B`:

1. Stealer reads `t=0, b=2`, pauses before CAS.
2. Owner `PushBottom(C)` → size becomes 2 (full), triggers `Grow(2, 0)` → `newBuf` cap=4 with `A@0, B@1`, then writes `newBuf[2]=C`, `_bottom=3`.
3. Stealer's CAS succeeds, `_top=1`.
4. Owner `PushBottom(D)` → `newBuf[3]=D`, `_bottom=4`.
5. Owner `PushBottom(E)` → `newBuf[4 & 3] = newBuf[0] = E`, **overwriting `A`**.
6. Stealer reads `_buffer = newBuf`, returns `newBuf[0 & 3] = E`.

The stealer claimed logical index 0 (which was `A`) but returns `E`. Worse, `A` is now lost — neither owner nor any stealer will ever consume it (no slot in the deque references it). This is exactly the "garbage" outcome the reviewer attributed to the original code, but it's actually the behavior of the proposed fix.

The original code is immune because `oldBuf[0]` is frozen as `A` after the swap and is never overwritten by any concurrent operation.

### Canonical references

Both reference Chase-Lev implementations I checked read the buffer **before** the CAS, identical to the original code:

- **Conor Williams' `riften::deque`** (C++17, based on Lê et al. PPoPP'13): `T x = _buffer.load(consume)->load(t);` then `compare_exchange_strong(...)`. Inline source comment: *"Must load before acquiring the slot as slot may be overwritten immediately after acquiring."*
- **Molecular Matters Job System 2.0** (production game-engine code): `Job* job = m_jobs[t & MASK];` then `_InterlockedCompareExchange(&m_top, t+1, t)`. Author's annotation: *"it is important to read the job before carrying out the CAS, because the location in the array could be overwritten by concurrent Push() operations happening after the CAS has completed."*

Sources:
- Chase & Lev (2005), *Dynamic Circular Work-Stealing Deque* — original paper
- Lê, Pop, Cohen, Zappa Nardelli (PPoPP'13), *Correct and Efficient Work-Stealing for Weak Memory Models* — modern memory-model-correct refinement: https://fzn.fr/readings/ppopp13.pdf
- https://github.com/ConorWilliams/ConcurrentDeque (riften/deque.hpp)
- https://blog.molecular-matters.com/2015/09/25/job-system-2-0-lock-free-work-stealing-part-3-going-lock-free/

### Action on the misleading comment

The comment `"Cache buffer and length to avoid race with Grow()"` is unhelpfully terse and led to confusion. I'll rewrite it to explicitly document **why** the read must happen before the CAS (citing the Push-after-CAS overwrite hazard above).

### Action on finding #6

The deterministic Grow-mid-Steal test the reviewer proposed is invalidated as written — with the original (correct) code, it would pass without any fix being needed (which is the point: there's no bug to deterministically reproduce). However, the reviewer's underlying concern — *"is the existing concurrent test wide enough to catch buffer-related races?"* — is fair. I'll add a **randomized stress test** that starts with a small initial capacity (forcing many concurrent Grows) and asserts every pushed item is consumed exactly once via `HashSet` aggregation across many iterations. This hardens coverage of the `Grow`/`Steal` interleaving without requiring a (non-existent) bug to reproduce.

---

Will push fixes for #2, #3, #4, #5 plus the substitute stress test for #6, and update the misleading comment in `Steal()` shortly.

quabug:
@opencara review

PR #52, opened 2026-04-23
[dev] Add Paradise.Rendering.WebGPU backend (M0b clear-color)

### LINKED ISSUE #41: Renderer M0b: WebGPU backend bootstrap (instance, device, surface, clear-color present)

quabug:
Closes #41

## Summary

Stand up the WebGPU (Dawn) backend on top of the `Paradise.Rendering` data contract from #40: instance, adapter, device, queue, native surface, and a clear-color present loop. The boundary established in #40 holds — `Paradise.Rendering.WebGPU` references **only** `WebGPUSharp` and `Paradise.Rendering`, and `Paradise.Rendering.Sample` is the **only** project that references `ppy.SDL3-CS`.

### Public surface (`Paradise.Rendering.WebGPU`)

- `WebGpuRenderer` — surface-backed ctor (`WebGpuRenderer(in SurfaceDescriptor)`) and `CreateHeadless(width, height)` static, plus `Resize(uint, uint)`, `RenderClearFrame(in ColorRgba)`, `Dispose()`. Headless is rejected on the surface ctor with a clear `ArgumentException`; the static factory is the only path.

### Internal state

- `WebGpuDevice` — owns `Instance` / `Adapter` / `Device` / `Queue`. Adapter request takes either a compatible surface or null (headless). `RequestAdapterSync` / `RequestDeviceSync` with 10s timeout.
- `SurfaceState` — owns the configured surface, picks `caps.Formats[0]` (or BGRA8Unorm fallback), reconfigures on `Resize`, unconfigures on dispose.
- `SurfaceFactory` — OS-dispatches on `SurfacePlatform`: Win32→`SurfaceSourceWindowsHWNDFFI`, Xlib→`SurfaceSourceXlibWindowFFI`, Wayland→`SurfaceSourceWaylandSurfaceFFI`, Cocoa→`SurfaceSourceMetalLayerFFI`. Headless is rejected here too.
- `SlotTable<T>` — generation-tracked slot allocator (free-list backed). Generation 0 is reserved for the invalid sentinel matching the public handle structs. Scaffold for M1 resource slots.

### Sample (`Paradise.Rendering.Sample`)

- `net10.0`, `<PublishAot>true</PublishAot>` (matches BT.Sample / ECS.Sample). Default mode: SDL3 windowed, resizable, presents `ColorRgba.CornflowerBlue` clear frames. `--headless N` mode: sets `SDL_HINT_VIDEO_DRIVER=dummy` via `SDL_SetHint` (managed `Environment.SetEnvironmentVariable` doesn't reliably reach SDL's libc `getenv` cache), constructs `WebGpuRenderer.CreateHeadless(...)`, runs N clear frames against an offscreen target, exits 0.
- OS-conditional surface mapping in the sample only — Win32 reads HWND, macOS reads NSWindow*, Linux reads Wayland display+surface (preferred) or X11 display+window. Cocoa support is documented as a known constraint per the issue spec; full CAMetalLayer wrap-up is deferred until M0b runs on macOS.

### Packages added

- `WebGPUSharp 0.5.2` — Dawn natives shipped in-package for win/osx/linux x64+arm64. `[DisableRuntimeMarshalling]`, no reflection.
- `ppy.SDL3-CS 2026.320.0` — pure P/Invoke, AOT-clean, native SDL3 per RID (also covers ios/android/win-x86).

Both versions live in `src/Directory.Packages.props`. The sample project is the only consumer of `ppy.SDL3-CS`; `Paradise.Rendering.WebGPU` does not depend on it.

## Test plan

- [x] Solution builds with `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` — 0 errors.
- [x] Full test suite: `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false --output normal` — **1953 passed, 0 failed, 0 skipped**. Four new tests in `Paradise.Rendering.WebGPU.Test` exercise the headless adapter path.
- [x] AOT publish: `dotnet publish src/Paradise.Rendering.Sample -c Release` — clean, no trim/AOT warnings on `linux-x64`. Other RIDs (`win-x64`, `osx-arm64`) covered by the published WebGPUSharp + ppy.SDL3-CS native bundles.
- [x] AOT smoke: `./Paradise.Rendering.Sample --headless 5` from the publish output prints `Headless mode: rendered 5 clear frames against an offscreen target.` and exits 0.
- [x] Public surface check: zero `IRenderDevice` / `IBuffer` / `ITexture` interfaces; only handles + descriptors + `WebGpuRenderer`.
- [x] Boundary check: `Paradise.Rendering.WebGPU` references only `WebGPUSharp` + `Paradise.Rendering`; `Paradise.Rendering.Sample` is the only project referencing `ppy.SDL3-CS`.

## Notes

- **`SDL_SetHint` vs `Environment.SetEnvironmentVariable`** — headless dummy-driver mode uses SDL's hint API (`SDL_SetHint(SDL_HINT_VIDEO_DRIVER, "dummy")`) rather than writing the process env var. Managed `Environment.SetEnvironmentVariable` does **not** reliably reach SDL's video subsystem because libc `getenv` may have cached at process start — once SDL's `SDL_Init` calls `getenv("SDL_VIDEODRIVER")` it can see stale/absent data even after the managed write. Please do **not** suggest reverting to `Environment.SetEnvironmentVariable`; the SDL hint API is the correct path. There's an in-line comment in `Program.cs` at the call site with the same rationale.
- **Headless adapter coverage** — the 1953-test run covers this PR on a Linux host with `mesa-vulkan-drivers` + `libvulkan1` (lavapipe → Dawn Vulkan adapter). On hosts without any Vulkan/Metal/DX12 adapter, the four smoke tests skip cleanly via `Skip.Test(...)` rather than failing; the AOT publish itself is the load-bearing M0 acceptance signal per the issue spec.
- **AOT publish** — verified on `linux-x64` from this branch; `win-x64` / `osx-arm64` / `osx-x64` / `win-arm64` / `linux-arm64` are covered by the published WebGPUSharp + ppy.SDL3-CS native bundles but not run end-to-end here. CI covers the other RIDs when the pipeline exists.
- **Pre-existing build noise** — `NETSDK1188 locale is not recognized` warnings (`cs`, `de`, `es`, `pt-br`, `zh-hans`, etc.) come from `Microsoft.Testing.Platform 2.0.2` / `Microsoft.Testing.Platform.MSBuild 2.0.2`'s packaging and were present before this PR. Not introduced or modifiable from this repo — upstream issue.
- **macOS Cocoa constraint** — the sample passes the SDL3-exposed `NSWindow*` rather than a `CAMetalLayer*`. Full CAMetalLayer wrap-up (creation on the main thread + attachment to `NSWindow.contentView.layer`) is deferred until M0b is exercised on macOS hardware. Documented in the issue spec; not a blocker for the headless acceptance path.

### DIFF
--- a/ParadiseEngine.slnx
+++ b/ParadiseEngine.slnx
@@ -9,6 +9,7 @@
   <Folder Name="/samples/">
     <Project Path="src/Paradise.BT.Sample/Paradise.BT.Sample.csproj" />
     <Project Path="src/Paradise.ECS.Sample/Paradise.ECS.Sample.csproj" />
+    <Project Path="src/Paradise.Rendering.Sample/Paradise.Rendering.Sample.csproj" />
   </Folder>
   <Folder Name="/sources/">
     <Project Path="src/Paradise.BLOB/Paradise.BLOB.csproj" />
@@ -20,6 +21,7 @@
     <Project Path="src/Paradise.ECS.Tag/Paradise.ECS.Tag.csproj" />
     <Project Path="src/Paradise.ECS/Paradise.ECS.csproj" />
     <Project Path="src/Paradise.Rendering/Paradise.Rendering.csproj" />
+    <Project Path="src/Paradise.Rendering.WebGPU/Paradise.Rendering.WebGPU.csproj" />
   </Folder>
   <Folder Name="/tests/">
     <Project Path="src/Paradise.BLOB.Test/Paradise.BLOB.Test.csproj" />
@@ -31,5 +33,6 @@
     <Project Path="src/Paradise.ECS.Generators.Test/Paradise.ECS.Generators.Test.csproj" />
     <Project Path="src/Paradise.ECS.Test/Paradise.ECS.Test.csproj" />
     <Project Path="src/Paradise.Rendering.Test/Paradise.Rendering.Test.csproj" />
+    <Project Path="src/Paradise.Rendering.WebGPU.Test/Paradise.Rendering.WebGPU.Test.csproj" />
   </Folder>
 </Solution>
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -8,6 +8,9 @@
     <!-- Runtime helpers -->
     <PackageVersion Include="System.Runtime.CompilerServices.Unsafe" Version="6.1.2" />
     <PackageVersion Include="System.Text.Json" Version="9.0.0" />
+    <!-- Rendering -->
+    <PackageVersion Include="WebGPUSharp" Version="0.5.2" />
+    <PackageVersion Include="ppy.SDL3-CS" Version="2026.320.0" />
     <!-- Testing (TUnit with Native AOT support) -->
     <PackageVersion Include="TUnit" Version="1.11.16" />
     <PackageVersion Include="Microsoft.Coyote.Test" Version="1.7.11" />
--- a/src/Paradise.Rendering.Sample/Paradise.Rendering.Sample.csproj
+++ b/src/Paradise.Rendering.Sample/Paradise.Rendering.Sample.csproj
@@ -1,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <IsPackable>false</IsPackable>
+    <PublishAot>true</PublishAot>
+    <RootNamespace>Paradise.Rendering.Sample</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../Paradise.Rendering/Paradise.Rendering.csproj" />
+    <ProjectReference Include="../Paradise.Rendering.WebGPU/Paradise.Rendering.WebGPU.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="ppy.SDL3-CS" />
+  </ItemGroup>
+
+</Project>
--- a/src/Paradise.Rendering.Sample/Program.cs
+++ b/src/Paradise.Rendering.Sample/Program.cs
@@ -1,0 +1,177 @@
+using System;
+using System.Runtime.InteropServices;
+using Paradise.Rendering;
+using Paradise.Rendering.WebGPU;
+using static SDL.SDL3;
+using SDL;
+
+namespace Paradise.Rendering.Sample;
+
+internal static class Program
+{
+    private const int InitialWidth = 640;
+    private const int InitialHeight = 480;
+
+    private static int Main(string[] args)
+    {
+        var headlessFrames = ParseHeadless(args);
+        try
+        {
+            return headlessFrames is int n ? RunHeadless(n) : RunWindowed();
+        }
+        catch (Exception ex)
+        {
+            Console.Error.WriteLine($"Sample failed: {ex}");
+            return 1;
+        }
+    }
+
+    private static int? ParseHeadless(string[] args)
+    {
+        for (var i = 0; i < args.Length; i++)
+        {
+            if (args[i] != "--headless") continue;
+            if (i + 1 >= args.Length || !int.TryParse(args[i + 1], out var n) || n <= 0)
+            {
+                Console.Error.WriteLine("Usage: --headless <positive integer>");
+                return -1;
+            }
+            return n;
+        }
+        return null;
+    }
+
+    private static int RunHeadless(int frameCount)
+    {
+        if (frameCount < 0) return 1;
+
+        // SDL still needs to initialize cleanly even though no window is opened — the dummy
+        // video driver lets it succeed in headless CI containers without a display server.
+        // Using SDL_SetHint instead of Environment.SetEnvironmentVariable: native libc getenv()
+        // can cache values from process start, so the managed env var doesn't reliably reach
+        // SDL's video subsystem; SDL_SetHint writes through the SDL hint API directly.
+        SDL_SetHint(SDL_HINT_VIDEO_DRIVER, "dummy"u8);
+
+        if (!SDL_Init(SDL_InitFlags.SDL_INIT_VIDEO))
+        {
+            Console.Error.WriteLine($"SDL_Init failed: {SDL_GetError()}");
+            return 1;
+        }
+
+        try
+        {
+            using var renderer = WebGpuRenderer.CreateHeadless(InitialWidth, InitialHeight);
+            for (var i = 0; i < frameCount; i++)
+                renderer.RenderClearFrame(ColorRgba.CornflowerBlue);
+            Console.WriteLine($"Headless mode: rendered {frameCount} clear frames against an offscreen target.");
+            return 0;
+        }
+        finally
+        {
+            SDL_Quit();
+        }
+    }
+
+    private static unsafe int RunWindowed()
+    {
+        if (!SDL_Init(SDL_InitFlags.SDL_INIT_VIDEO))
+        {
+            Console.Error.WriteLine($"SDL_Init failed: {SDL_GetError()}");
+            return 1;
+        }
+
+        SDL_Window* window = null;
+        WebGpuRenderer? renderer = null;
+        try
+        {
+            window = SDL_CreateWindow("Paradise.Rendering — Clear Color", InitialWidth, InitialHeight, SDL_WindowFlags.SDL_WINDOW_RESIZABLE);
+            if (window == null)
+            {
+                Console.Error.WriteLine($"SDL_CreateWindow failed: {SDL_GetError()}");
+                return 1;
+            }
+
+            var surfaceDesc = BuildSurfaceDescriptor(window);
+            renderer = new WebGpuRenderer(in surfaceDesc);
+
+            var quit = false;
+            SDL_Event ev;
+            while (!quit)
+            {
+                while (SDL_PollEvent(&ev))
+                {
+                    var type = (SDL_EventType)ev.type;
+                    if (type == SDL_EventType.SDL_EVENT_QUIT)
+                    {
+                        quit = true;
+                    }
+                    else if (type == SDL_EventType.SDL_EVENT_WINDOW_CLOSE_REQUESTED)
+                    {
+                        quit = true;
+                    }
+                    else if (type == SDL_EventType.SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED ||
+                             type == SDL_EventType.SDL_EVENT_WINDOW_RESIZED)
+                    {
+                        var w = ev.window.data1;
+                        var h = ev.window.data2;
+                        if (w > 0 && h > 0)
+                            renderer.Resize((uint)w, (uint)h);
+                    }
+                }
+                renderer.RenderClearFrame(ColorRgba.CornflowerBlue);
+            }
+
+            return 0;
+        }
+        finally
+        {
+            renderer?.Dispose();
+            if (window != null) SDL_DestroyWindow(window);
+            SDL_Quit();
+        }
+    }
+
+    private static unsafe SurfaceDescriptor BuildSurfaceDescriptor(SDL_Window* window)
+    {
+        var props = SDL_GetWindowProperties(window);
+
+        int w = 0, h = 0;
+        SDL_GetWindowSizeInPixels(window, &w, &h);
+        var width = (uint)Math.Max(1, w);
+        var height = (uint)Math.Max(1, h);
+
+        if (OperatingSystem.IsWindows())
+        {
+            var hwnd = SDL_GetPointerProperty(props, SDL_PROP_WINDOW_WIN32_HWND_POINTER, IntPtr.Zero);
+            return new SurfaceDescriptor(SurfacePlatform.Win32, IntPtr.Zero, hwnd, width, height);
+        }
+
+        if (OperatingSystem.IsMacOS())
+        {
+            // SDL3 exposes the NSWindow*, not a CAMetalLayer*. The minimal sample passes the
+            // window pointer; full Cocoa hookup (creating CAMetalLayer on the main thread and
+            // attaching it to the content view) is documented as a known macOS constraint in
+            // the issue spec and is left to be exercised when M0b runs on macOS.
+            var nsWindow = SDL_GetPointerProperty(props, SDL_PROP_WINDOW_COCOA_WINDOW_POINTER, IntPtr.Zero);
+            return new SurfaceDescriptor(SurfacePlatform.Cocoa, IntPtr.Zero, nsWindow, width, height);
+        }
+
+        if (OperatingSystem.IsLinux())
+        {
+            var wlDisplay = SDL_GetPointerProperty(props, SDL_PROP_WINDOW_WAYLAND_DISPLAY_POINTER, IntPtr.Zero);
+            if (wlDisplay != IntPtr.Zero)
+            {
+                var wlSurface = SDL_GetPointerProperty(props, SDL_PROP_WINDOW_WAYLAND_SURFACE_POINTER, IntPtr.Zero);
+                return new SurfaceDescriptor(SurfacePlatform.Wayland, wlDisplay, wlSurface, width, height);
+            }
+
+            var x11Display = SDL_GetPointerProperty(props, SDL_PROP_WINDOW_X11_DISPLAY_POINTER, IntPtr.Zero);
+            var x11Window = SDL_GetNumberProperty(props, SDL_PROP_WINDOW_X11_WINDOW_NUMBER, 0);
+            return new SurfaceDescriptor(SurfacePlatform.Xlib, x11Display, (IntPtr)x11Window, width, height);
+        }
+
+        throw new PlatformNotSupportedException(
+            $"Surface mapping for the current OS ({RuntimeInformation.OSDescription}) is not implemented; " +
+            "use --headless on this platform.");
+    }
+}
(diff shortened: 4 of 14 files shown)


quabug:
@opencara review

quabug:
## OpenCara round-1 fixes (commit a9a87be)

All three findings addressed:

**1. [major] Surface-loss reconfigure no-op** — added `SurfaceState.Reconfigure()` (unconditional `Configure()` call) and routed the `Outdated`/`Lost` branch in `WebGpuRenderer.RenderClearFrame` through it. The `Resize` short-circuit on equal dimensions stays — that's correct for the resize path, but recovery now uses a dedicated entrypoint that always re-applies the configuration.

**2. [major] macOS Cocoa NSWindow vs CAMetalLayer mismatch** — replaced the silent `NSWindow*`-as-`CAMetalLayer*` handoff with an explicit `PlatformNotSupportedException` that tells users to use `--headless N` on macOS until full CAMetalLayer hookup lands. Per OpenCara's "reject windowed macOS at runtime until that hookup exists" recommendation. Headless adapter path on macOS is unaffected.

**3. [minor] Tests skipped device-creation failures** — introduced `AdapterUnavailableException` (public, in `Paradise.Rendering.WebGPU`) thrown only when the adapter request itself returns null. Narrowed the test `catch` from `InvalidOperationException` + `Contains("adapter")` filter to a typed `catch (AdapterUnavailableException ex)`. Also tightened the device-create error message to "WebGPU device creation failed." so it no longer mentions the word "adapter" — even string-matching consumers won't conflate the two.

Local verification re-run after the fixes:
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` → 0 errors
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false --output normal` → **1953/1953 passed, 0 failed, 0 skipped**
- `dotnet publish src/Paradise.Rendering.Sample -c Release` → clean AOT
- `./Paradise.Rendering.Sample --headless 5` → "rendered 5 clear frames against an offscreen target." exit 0

quabug:
@opencara review

quabug:
## Round-2 reply (commit e4b4469)

Three reviewers approved (claude-sonnet-4-6, claude-opus-4-7, glm-5/qwen). Per the development-workflow gate ("Repeat until no critical/major issues remain") the remaining items are all [minor]; addressing the cheapest one and noting the rest as out-of-scope-for-M0b.

**Applied:**
- **Review 1 question — `current.Texture!` null safety** — replaced the null-forgiving operator with an explicit `?? throw new InvalidOperationException` that names the surface status. If a future WebGPUSharp version violates the post-success-status invariant, the diagnostic surfaces in the message rather than as a bare `NullReferenceException`.

**Declined (with evidence):**
- **Review 1 [minor] proven defect — per-frame `TextureView`/`CommandEncoder`/`CommandBuffer` not disposed.** The recommendation cannot be applied as written: I inspected `WebGPUSharp 0.5.2`'s public surface and confirmed via reflection that `WebGpuSharp.TextureView`, `WebGpuSharp.CommandEncoder`, `WebGpuSharp.CommandBuffer`, and `WebGpuSharp.RenderPassEncoder` expose no public `Dispose`, `Destroy`, or `Release` method (only `WebGpuSharp.Texture` exposes `Destroy()`, which is already used for the offscreen target on resize/dispose). The library's documented release model for these specific wrappers is finalizer-only — the underlying FFI handle layer (`WebGpuSharp.FFI.TextureViewHandle`, etc.) does have `Release()`/`Dispose()`, but reaching past the safe-wrapper boundary is a scope expansion beyond M0b's clear-color present loop. The structural concern is real for long-running workloads; it's tracked implicitly by the M1 command-stream redesign (which will need its own buffer-recycling strategy regardless of how the M0 single-frame path handles wrapper lifetime).

**Out of scope for this PR (Review 2 risks):**
- `Lost` semantics differentiation from `Outdated` (device-lost vs swapchain-lost) — for M0b's clear-color scope the bundled-recovery path is acceptable; correct device-loss recovery requires renderer reconstruction which is M1+ scaffolding.
- Construction-failure cleanup of `instance`/`nativeSurface` wrappers — same finalizer-only model as above.

Local re-verification:
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` → 0 errors
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false --output normal` → **1953/1953 passed, 0 failed, 0 skipped**

Proceeding to squash merge per the development-workflow Step 3 gate (no critical/major issues remaining).

Issue #54, opened 2026-04-23
Renderer M0b follow-up: differentiate Lost vs Outdated surface semantics + construction-failure cleanup


quabug:
## Why

PR #52 (squash `2be1507`, merged 2026-04-23) round-2 OpenCara review flagged two related risks that dev-41 deferred to M1:

1. **`Lost` vs `Outdated` semantics** — current code routes both through the same `Reconfigure()` path, but the WebGPU spec gives them different meanings. `Outdated` means "swapchain dimensions/format don't match surface" (cheap reconfigure), `Lost` means "device or surface is fundamentally gone" (may require full device recreation, not just reconfigure). Treating them identically silently masks device-loss scenarios that warrant escalation.
2. **Construction-failure cleanup** — if `WebGpuRenderer` ctor throws between resource allocations (e.g. surface created but device-init fails), partially-allocated native handles aren't released. Single-frame leak on init failure; not a recurring leak but worth fixing for correctness.

Both deferred to M1 (#42) since M0b's clear-color path doesn't exercise device-loss in practice.

### For Lost vs Outdated

Split the swapchain-state branch:
- `Outdated` → existing `SurfaceState.Reconfigure()` (cheap)
- `Lost` → new `WebGpuRenderer.HandleDeviceLost()` path that:
  - Releases all currently-cached resources
  - Requests a new device from the same instance/adapter (or new adapter if old is gone)
  - Reconfigures surface against the new device
  - Restores any persistent state the application expects
  - If recovery fails after N attempts, throws / surfaces an error to the caller

WebGPUSharp exposes a device-lost callback (`Device.SetDeviceLostCallback` or similar — verify against 0.5.2 surface). Wire that as the trigger for `HandleDeviceLost` rather than waiting until next frame's surface-acquire.

### For construction-failure cleanup

Wrap `WebGpuRenderer` ctor body in try/catch:
- On exception, release any partially-allocated handles (instance, adapter, device, surface)
- Re-throw with the original exception preserved
- Add a regression test that forces a controlled init failure (mock or test seam) and asserts no native-resource leak (slot table occupancy = 0 post-throw)

## Acceptance criteria

- `Lost` and `Outdated` handled distinctly; device-lost path tested via WebGPUSharp callback (or programmatic forcing if available, otherwise documented as manually-verified)
- Construction failure releases all partial allocations; regression test in place
- No regression in M0b headless smoke or AOT publish

## References

[...]

Issue #53, opened 2026-04-23
Renderer M0b follow-up: WebGPUSharp per-frame TextureView/CommandEncoder/CommandBuffer disposal


quabug:
## Why

PR #52 (squash `2be1507`, merged 2026-04-23) round-2 OpenCara review surfaced a finding that per-frame `TextureView`, `CommandEncoder`, and `CommandBuffer` instances are not disposed in `WebGpuRenderer`. dev-41 declined the immediate fix with API-surface evidence: WebGPUSharp 0.5.2's safe wrappers expose **no public Dispose/Destroy/Release** for these three types — only `Texture` has `Destroy()` (already used). The library's release model for these wrappers is finalizer-only by design, so the runtime impact is bounded by GC pressure rather than leaks per se.

Acceptable for M0b clear-color (single draw + present per frame, low allocation pressure) but worth revisiting once M1 (#42) introduces real draw paths and per-frame allocations climb.

## What

Investigate options:
1. **Upstream PR to WebGPUSharp** adding `IDisposable` to `TextureView` / `CommandEncoder` / `CommandBuffer` safe wrappers. Cleanest fix; benefits the whole ecosystem.
2. **Local thin wrapper layer** — wrap the three types in a `PooledTextureView` / `PooledCommandEncoder` / etc. that holds the WebGPUSharp instance + a per-frame deferred-release queue. Trade-off: extra indirection, but works without upstream changes.
3. **Reach into the unsafe FFI** and call `wgpuTextureViewRelease` / etc. directly. Workable but couples us tightly to WebGPUSharp internals; brittle.

Recommend **Option 1 first** (file an upstream issue/PR; bus factor of one means responsiveness is uncertain but worth trying), with **Option 2 as fallback** if upstream is unresponsive after ~2 weeks.

## Acceptance criteria

- Per-frame `TextureView` / `CommandEncoder` / `CommandBuffer` instances released deterministically (not finalizer-only)
- No regression in headless smoke or AOT publish
- If Option 1: link to upstream PR; if Option 2: new wrapper types are AOT-clean and have round-trip tests

## References

- PR #52: https://github.com/ParadiseEngine/ParadiseEngine/pull/52
- WebGPUSharp 0.5.2 source — verify the safe-wrapper Dispose contract before starting; if the upstream surface has changed, this issue may already be obsolete
- dev-41's rebuttal comment in PR #52 round-2 review thread (cite the API-evidence findings)

## Coordination

[...]

Issue #43, opened 2026-04-22
Renderer M2: textures, samplers, uniforms, depth (textured indexed mesh)


quabug:
Parent: #39 (M2 — basic scene resources)
Depends on: #42 (M1 draw path must be working)

**Stack** (inherited from #41): WebGPUSharp (Dawn natives bundled) + ppy.SDL3-CS. No Silk.NET dependencies.

**Shader toolchain** (inherited from #42): Slang v2026.7 → WGSL, build-time, vendored. Reflection JSON is now exercised on the **non-trivial** binding path here.

## Why

M1 proves the boundary handles a single colored triangle and the Slang→WGSL build pipeline works end-to-end. M2 expands it to cover what every real shader needs: textures, samplers, depth, and uniform/storage buffers. After this, M3 (ECS extraction) can produce real per-frame data without the renderer growing more API surface.

This milestone also stresses the Slang reflection path: the `material.slang` shader declares group 0 = uniform buffer + group 1 = sampled texture + sampler. The reflection JSON drives `BindGroupLayoutDescriptor` construction so we catch any Slang WGSL backend gaps (e.g. upstream issues #5222 / #5612) on the non-trivial binding case before #44 builds on top.

End state: the sample renders a textured, indexed quad with a depth buffer and a per-frame uniform (e.g., projection matrix or `float time`), all driven by a `.slang` shader and its reflection JSON.

### Extend `Paradise.Rendering` data contracts

The `BindGroupLayoutDesc` / `BindGroupLayoutEntryDesc` / `PipelineLayoutDesc` / `BindingResourceType` types already live in `Paradise.Rendering` as records (landed in #40 to lock the Slang-reflection-shaped contract). M2 adds only the **runtime** companions:

```csharp
public readonly struct BindGroupLayoutHandle : IEquatable<BindGroupLayoutHandle> { ... } // 16-byte handle, same shape as M0a handles
public readonly struct BindGroupHandle       : IEquatable<BindGroupHandle> { ... }

public readonly struct BindGroupBufferEntry  { public uint Binding; public BufferHandle Buffer; public ulong Offset; public ulong Size; }
public readonly struct BindGroupTextureEntry { public uint Binding; public RenderViewHandle View; }
public readonly struct BindGroupSamplerEntry { public uint Binding; public SamplerHandle Sampler; }

public readonly struct BindGroupDesc
{
    public string? Name;
    public BindGroupLayoutHandle Layout;
[...]

PR #55, opened 2026-04-23
[dev] Add Slang->WGSL toolchain + triangle draw path (M1)

### LINKED ISSUE #42: Renderer M1: minimal draw path (shaders, pipeline, vertex/index buffers, triangle)

quabug:
Closes #42

## Summary

- Build-time Slang->WGSL toolchain with vendored slangc 2026.7 (`tools/slang/slang.manifest.json`, `src/Slang.targets`, `tools/slang/SlangBootstrap.cs`). RestoreSlang downloads + SHA256-verifies + caches under `$(NuGetPackageRoot)_slang/$(SlangVersion)/$(RuntimeIdentifier)`. CompileSlangShaders globs `Shaders/**/*.slang` and emits `<name>.wgsl` + `<name>.reflection.json` as `<EmbeddedResource>`s.
- `Paradise.Rendering` data contracts: filled-in `PipelineDesc` (cache-keyable via `ContentHash()`), 40-byte explicit-layout discriminated-union `RenderCommand`, `RenderCommandStream` + `RenderCommandEncoder` ref struct over `IBufferWriter<RenderCommand>`, `VertexFormats.ByteSize` helper.
- `Paradise.Rendering.WebGPU` backend: per-resource `SlotTable<T>`, `Create*`/`Destroy*` for shaders / buffers / pipelines, `CreatePipeline(ShaderProgramDesc, TextureFormat)` overload, `Submit(RenderCommandStream)`, `PipelineCache`, `DeferredDestructionQueue`, `StaleHandleException`. `ShaderProgramLoader` (internal) parses raw slangc reflection-JSON output and constructs `ShaderProgramDesc` programmatically — vertex layout derived from `entryPoints[].parameters[].fields` walking, never hand-coded.
- Sample: `Shaders/triangle.slang` (HLSL-style vs/fs), `TriangleScene` builds the pipeline from the reflection record, uploads vertices once, and submits a single Draw command per frame. Headless `--headless N` mode and the AOT-published binary both render the triangle.
- Tests: 52 new — encoder round-trip, `PipelineDesc` content hash determinism, slot table generation bumping (use-after-free guard), pipeline cache hit on identical content, deferred destruction frame timing, reflection round-trip from embedded `triangle.reflection.json` (vertex layout matches the .slang declaration).
- CI: `actions/cache` step keyed on `tools/slang/slang.manifest.json` SHA, caching `~/.nuget/packages/_slang`.

## Test plan

- [x] `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` — clean (only NETSDK1188 locale warnings from external test packages, not errors)
- [x] `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false --output normal` — 1971/1971 pass
- [x] `dotnet publish src/Paradise.Rendering.Sample -c Release` — AOT publish clean
- [x] `bin/Release/net10.0/linux-x64/publish/Paradise.Rendering.Sample --headless 5` — prints "Headless mode: rendered 5 triangle frames against an offscreen target."
- [x] Slang restore is idempotent: second `dotnet build` reuses `$(SlangDir)/.installed` marker, no re-download

## Notes

- Slang reflection-JSON schema does not match the `ShaderProgramDesc` records from #40 1:1 (the records were a simplified shape; slangc's actual output is more nested). The loader (`Internal/ShaderProgramLoader.cs` + `Internal/SlangReflectionSchema.cs`) absorbs the gap by parsing the raw schema and building `ShaderProgramDesc` programmatically. The #40 records and their source-gen JSON context are unchanged; the existing `reflection.sample.json` fixture in `Paradise.Rendering.Test` continues to exercise the engine-canonical record round-trip.
- Linux toolchain pin: `linux-x86_64-glibc-2.27` archive (matches the documented Ubuntu 22.04+ floor; runs cleanly on `ubuntu-latest` 24.04).
- `WebGpuRenderer.LoadShaderProgram` is exposed as a `public static` helper that wraps the internal loader so the sample (and downstream consumers without `InternalsVisibleTo`) can use it without leaking the loader type.
- M1 backend convention: a single color attachment per `RenderPassDesc` always targets the backbuffer view; multi-attachment / offscreen-target rendering throws `NotSupportedException` and is deferred to M2 (#43). `SetBindGroup` is encoded but no-op'd by the backend (also M2).
- Per-frame native disposal (`TextureView` / `CommandEncoder` / `CommandBuffer`) inherits the WebGPUSharp 0.5.2 finalizer model from #41; #53 tracks the upstream/wrapper fix.
- `Lost` and `Outdated` surface-acquire statuses are still routed through the same reconfigure path; #54 tracks the split.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -22,6 +22,17 @@ jobs:
         with:
           global-json-file: global.json
 
+      # Cache the downloaded slangc toolchain across runs. Keyed on the manifest SHA so a Slang
+      # version bump invalidates the cache automatically; the cache root mirrors RestoreSlang's
+      # default ($(NuGetPackageRoot)_slang) which lives under ~/.nuget/packages on Linux runners.
+      - name: Cache slangc
+        uses: actions/cache@v4
+        with:
+          path: ~/.nuget/packages/_slang
+          key: slang-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('tools/slang/slang.manifest.json') }}
+          restore-keys: |
+            slang-${{ runner.os }}-${{ runner.arch }}-
+
       - name: Prepare CI solution
         run: |
           cp ParadiseEngine.slnx ParadiseEngine.ci.slnx
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,10 @@
 [Oo]ut/
 [Pp]ublish/
 
+# Slang toolchain cache (RestoreSlang fallback path when NuGetPackageRoot is unset).
+# Production cache lives under ~/.nuget/packages/_slang and is shared across builds.
+.slang-cache/
+
 # IDE
 .vs/
 .vscode/
--- a/src/Paradise.Rendering.Sample/Paradise.Rendering.Sample.csproj
+++ b/src/Paradise.Rendering.Sample/Paradise.Rendering.Sample.csproj
@@ -17,4 +17,10 @@
     <PackageReference Include="ppy.SDL3-CS" />
   </ItemGroup>
 
+  <Import Project="../Slang.targets" />
+
+  <ItemGroup>
+    <SlangShader Include="Shaders/**/*.slang" />
+  </ItemGroup>
+
 </Project>
--- a/src/Paradise.Rendering.Sample/Program.cs
+++ b/src/Paradise.Rendering.Sample/Program.cs
@@ -61,9 +61,10 @@ internal static class Program
         try
         {
             using var renderer = WebGpuRenderer.CreateHeadless(InitialWidth, InitialHeight);
+            using var scene = new TriangleScene(renderer);
             for (var i = 0; i < frameCount; i++)
-                renderer.RenderClearFrame(ColorRgba.CornflowerBlue);
-            Console.WriteLine($"Headless mode: rendered {frameCount} clear frames against an offscreen target.");
+                scene.RenderFrame();
+            Console.WriteLine($"Headless mode: rendered {frameCount} triangle frames against an offscreen target.");
             return 0;
         }
         finally
@@ -93,6 +94,7 @@ internal static class Program
 
             var surfaceDesc = BuildSurfaceDescriptor(window);
             renderer = new WebGpuRenderer(in surfaceDesc);
+            using var scene = new TriangleScene(renderer);
 
             var quit = false;
             SDL_Event ev;
@@ -118,7 +120,7 @@ internal static class Program
                             renderer.Resize((uint)w, (uint)h);
                     }
                 }
-                renderer.RenderClearFrame(ColorRgba.CornflowerBlue);
+                scene.RenderFrame();
             }
 
             return 0;
--- a/src/Paradise.Rendering.Sample/Shaders/triangle.slang
+++ b/src/Paradise.Rendering.Sample/Shaders/triangle.slang
@@ -1,0 +1,30 @@
+// Authoring source-of-truth for the M1 triangle. Compiled to triangle.wgsl + triangle.reflection.json
+// at build time by src/Slang.targets. The runtime sample loads the embedded outputs and derives the
+// vertex layout from the reflection record (NOT hand-coded) — that's the M1 design contract.
+
+struct VsIn
+{
+    float2 pos : POSITION;
+    float3 color : COLOR;
+};
+
+struct VsOut
+{
+    float4 pos : SV_Position;
+    float3 color : COLOR;
+};
+
+[shader("vertex")]
+VsOut vs_main(VsIn input)
+{
+    VsOut o;
+    o.pos = float4(input.pos, 0.0, 1.0);
+    o.color = input.color;
+    return o;
+}
+
+[shader("fragment")]
+float4 fs_main(VsOut input) : SV_Target
+{
+    return float4(input.color, 1.0);
+}
--- a/src/Paradise.Rendering.Sample/TriangleScene.cs
+++ b/src/Paradise.Rendering.Sample/TriangleScene.cs
@@ -1,0 +1,72 @@
+using System;
+using System.Buffers;
+using System.Reflection;
+using Paradise.Rendering;
+using Paradise.Rendering.WebGPU;
+
+namespace Paradise.Rendering.Sample;
+
+/// <summary>One-frame draw plumbing for the M1 triangle. Loads the embedded Slang outputs, builds a
+/// pipeline whose vertex layout comes from the reflection record (NOT hand-coded), uploads the
+/// vertex buffer once, and produces a <see cref="RenderCommandStream"/> per frame.</summary>
+internal sealed class TriangleScene : IDisposable
+{
+    // Three positions + three colors interleaved (xy, rgb): centered triangle, primary colors at
+    // each corner. Layout MUST match what triangle.slang declares for VsIn — the engine asserts
+    // this in the reflection round-trip test, so any drift here would surface immediately.
+    private static readonly float[] s_vertices =
+    {
+        // pos.x, pos.y, color.r, color.g, color.b
+         0.0f,  0.6f,   1.0f, 0.0f, 0.0f,
+        -0.6f, -0.6f,   0.0f, 1.0f, 0.0f,
+         0.6f, -0.6f,   0.0f, 0.0f, 1.0f,
+    };
+
+    private readonly WebGpuRenderer _renderer;
+    private readonly BufferHandle _vertexBuffer;
+    private readonly PipelineHandle _pipeline;
+    private readonly RenderPassDesc[] _passes;
+
+    public TriangleScene(WebGpuRenderer renderer)
+    {
+        _renderer = renderer;
+
+        var program = WebGpuRenderer.LoadShaderProgram(typeof(TriangleScene).Assembly, "Shaders.triangle");
+        _pipeline = renderer.CreatePipeline(program, renderer.ColorFormat);
+
+        var bufferDesc = new BufferDesc(
+            Name: "TriangleVertices",
+            Size: (ulong)(s_vertices.Length * sizeof(float)),
+            Usage: BufferUsage.Vertex,
+            MappedAtCreation: false);
+        _vertexBuffer = renderer.CreateBufferWithData(in bufferDesc, (ReadOnlySpan<float>)s_vertices);
+
+        _passes = new RenderPassDesc[1];
+        _passes[0] = new RenderPassDesc(colorAttachmentCount: 1);
+        _passes[0].Colors.Slot0 = new ColorAttachmentDesc(
+            View: RenderViewHandle.Invalid,  // backbuffer — see WebGpuRenderer.BeginPass
+            Load: LoadOp.Clear,
+            Store: StoreOp.Store,
+            ClearValue: ColorRgba.CornflowerBlue);
+    }
+
+    public void RenderFrame()
+    {
+        var writer = new ArrayBufferWriter<RenderCommand>(8);
+        var encoder = new RenderCommandEncoder(writer);
+        encoder.BeginPass(0);
+        encoder.SetPipeline(_pipeline);
+        encoder.SetVertexBuffer(0, _vertexBuffer, 0, (ulong)(s_vertices.Length * sizeof(float)));
+        encoder.Draw(new DrawCommand(VertexCount: 3, InstanceCount: 1, FirstVertex: 0, FirstInstance: 0));
+        encoder.EndPass();
+
+        var stream = new RenderCommandStream(writer.WrittenMemory, _passes);
+        _renderer.Submit(in stream);
+    }
+
+    public void Dispose()
+    {
+        _renderer.DestroyPipeline(_pipeline);
+        _renderer.DestroyBuffer(_vertexBuffer);
+    }
+}
--- a/src/Paradise.Rendering.Sample/TriangleScene.cs
+++ b/src/Paradise.Rendering.Sample/TriangleScene.cs
@@ -40,8 +40,7 @@ internal sealed class TriangleScene : IDisposable
         var bufferDesc = new BufferDesc(
             Name: "TriangleVertices",
             Size: (ulong)(s_vertices.Length * sizeof(float)),
-            Usage: BufferUsage.Vertex,
-            MappedAtCreation: false);
+            Usage: BufferUsage.Vertex);
         _vertexBuffer = renderer.CreateBufferWithData(in bufferDesc, (ReadOnlySpan<float>)s_vertices);
 
         _passes = new RenderPassDesc[1];
--- a/src/Paradise.Rendering.Sample/TriangleScene.cs
+++ b/src/Paradise.Rendering.Sample/TriangleScene.cs
@@ -26,6 +26,9 @@ internal sealed class TriangleScene : IDisposable
     private readonly BufferHandle _vertexBuffer;
     private readonly PipelineHandle _pipeline;
     private readonly RenderPassDesc[] _passes;
+    // Reused per-frame so the sample's render loop is allocation-free at steady state. Cleared
+    // at the top of RenderFrame; capacity grows once and stays put across frames.
+    private readonly ArrayBufferWriter<RenderCommand> _commandWriter = new(8);
 
     public TriangleScene(WebGpuRenderer renderer)
     {
@@ -52,15 +55,15 @@ internal sealed class TriangleScene : IDisposable
 
     public void RenderFrame()
     {
-        var writer = new ArrayBufferWriter<RenderCommand>(8);
-        var encoder = new RenderCommandEncoder(writer);
+        _commandWriter.ResetWrittenCount();
+        var encoder = new RenderCommandEncoder(_commandWriter);
         encoder.BeginPass(0);
         encoder.SetPipeline(_pipeline);
         encoder.SetVertexBuffer(0, _vertexBuffer, 0, (ulong)(s_vertices.Length * sizeof(float)));
         encoder.Draw(new DrawCommand(VertexCount: 3, InstanceCount: 1, FirstVertex: 0, FirstInstance: 0));
         encoder.EndPass();
 
-        var stream = new RenderCommandStream(writer.WrittenMemory, _passes);
+        var stream = new RenderCommandStream(_commandWriter.WrittenMemory, _passes);
         _renderer.Submit(in stream);
     }
 
--- a/src/Paradise.Rendering.Test/PipelineDescTests.cs
+++ b/src/Paradise.Rendering.Test/PipelineDescTests.cs
@@ -84,4 +84,63 @@ public class PipelineDescTests
         await Assert.That(a == b).IsTrue();
         await Assert.That(a.GetHashCode()).IsEqualTo(b.GetHashCode());
     }
+
+    [Test]
+    public async Task layout_is_compared_structurally_not_by_reference()
+    {
+        // Regression: ShaderProgramLoader.BuildProgramDesc mints a fresh PipelineLayoutDesc per
+        // load, so two loads of the same shader produce distinct Layout references. Reference-
+        // equality Equals/ContentHash defeated the pipeline cache for that path — a cache miss
+        // retained a duplicate native pipeline for the renderer's lifetime. Structural equality
+        // (bind-group + push-constant deep walk) fixes the cache hit.
+        var layoutA = new PipelineLayoutDesc(
+            Groups: Array.Empty<BindGroupLayoutDesc>(),
+            PushConstants: Array.Empty<PushConstantRangeDesc>());
+        var layoutB = new PipelineLayoutDesc(
+            Groups: Array.Empty<BindGroupLayoutDesc>(),
+            PushConstants: Array.Empty<PushConstantRangeDesc>());
+        await Assert.That(ReferenceEquals(layoutA, layoutB)).IsFalse();
+
+        var a = BuildSample() with { };
+        a = new PipelineDesc
+        {
+            VertexShader = new ShaderHandle(1, 1),
+            VertexEntryPoint = "vs_main",
+            FragmentShader = new ShaderHandle(2, 1),
+            FragmentEntryPoint = "fs_main",
+            VertexLayouts = new[] { new VertexBufferLayoutDesc(20, VertexStepMode.Vertex, new[] { new VertexAttributeDesc(0, VertexFormat.Float32x2, 0), new VertexAttributeDesc(1, VertexFormat.Float32x3, 8) }) },
+            Topology = PrimitiveTopology.TriangleList,
+            StripIndexFormat = IndexFormat.Uint16,
+            ColorFormat = TextureFormat.Bgra8Unorm,
+            DepthStencilFormat = null,
+            Layout = layoutA,
+        };
+        var b = a with { Layout = layoutB };
+
+        await Assert.That(a).IsEqualTo(b);
+        await Assert.That(a.ContentHash()).IsEqualTo(b.ContentHash());
+    }
+
+    [Test]
+    public async Task layout_structural_inequality_is_detected()
+    {
+        var layoutEmpty = new PipelineLayoutDesc(
+            Groups: Array.Empty<BindGroupLayoutDesc>(),
+            PushConstants: Array.Empty<PushConstantRangeDesc>());
+        var layoutWithBinding = new PipelineLayoutDesc(
+            Groups: new[]
+            {
+                new BindGroupLayoutDesc(0, new[]
+                {
+                    new BindGroupLayoutEntryDesc(0, ShaderStage.Vertex, BindingResourceType.UniformBuffer, 16),
+                }),
+            },
+            PushConstants: Array.Empty<PushConstantRangeDesc>());
+
+        var a = BuildSample() with { Layout = layoutEmpty };
+        var b = a with { Layout = layoutWithBinding };
+
+        await Assert.That(a).IsNotEqualTo(b);
+        await Assert.That(a.ContentHash()).IsNotEqualTo(b.ContentHash());
+    }
 }
--- a/src/Paradise.Rendering.Test/PipelineDescTests.cs
+++ b/src/Paradise.Rendering.Test/PipelineDescTests.cs
@@ -1,0 +1,87 @@
+using System;
+
+namespace Paradise.Rendering.Test;
+
+public class PipelineDescTests
+{
+    private static PipelineDesc BuildSample(string? name = null, ShaderHandle? overrideVs = null)
+    {
+        var attrs = new[]
+        {
+            new VertexAttributeDesc(0, VertexFormat.Float32x2, 0),
+            new VertexAttributeDesc(1, VertexFormat.Float32x3, 8),
+        };
+        var layouts = new[]
+        {
+            new VertexBufferLayoutDesc(20, VertexStepMode.Vertex, attrs),
+        };
+        return new PipelineDesc
+        {
+            Name = name,
+            VertexShader = overrideVs ?? new ShaderHandle(1, 1),
+            VertexEntryPoint = "vs_main",
+            FragmentShader = new ShaderHandle(2, 1),
+            FragmentEntryPoint = "fs_main",
+            VertexLayouts = layouts,
+            Topology = PrimitiveTopology.TriangleList,
+            StripIndexFormat = IndexFormat.Uint16,
+            ColorFormat = TextureFormat.Bgra8Unorm,
+            DepthStencilFormat = null,
+            Layout = null,
+        };
+    }
+
+    [Test]
+    public async Task content_hash_is_stable_for_equal_descriptors()
+    {
+        var a = BuildSample("a");
+        var b = BuildSample("b"); // different name, same content
+        await Assert.That(a.ContentHash()).IsEqualTo(b.ContentHash());
+        await Assert.That(a).IsEqualTo(b);
+    }
+
+    [Test]
+    public async Task content_hash_differs_when_shader_handle_differs()
+    {
+        var a = BuildSample();
+        var b = BuildSample(overrideVs: new ShaderHandle(99, 1));
+        await Assert.That(a.ContentHash()).IsNotEqualTo(b.ContentHash());
+        await Assert.That(a).IsNotEqualTo(b);
+    }
+
+    [Test]
+    public async Task content_hash_differs_when_vertex_layout_differs()
+    {
+        var a = BuildSample();
+        var bAttrs = new[]
+        {
+            new VertexAttributeDesc(0, VertexFormat.Float32x4, 0),
+        };
+        var b = a with { }; // record-struct-style copy isn't supported on plain struct, rebuild
+        var bLayouts = new[] { new VertexBufferLayoutDesc(16, VertexStepMode.Vertex, bAttrs) };
+        b = new PipelineDesc
+        {
+            Name = a.Name,
+            VertexShader = a.VertexShader,
+            VertexEntryPoint = a.VertexEntryPoint,
+            FragmentShader = a.FragmentShader,
+            FragmentEntryPoint = a.FragmentEntryPoint,
+            VertexLayouts = bLayouts,
+            Topology = a.Topology,
+            StripIndexFormat = a.StripIndexFormat,
+            ColorFormat = a.ColorFormat,
+            DepthStencilFormat = a.DepthStencilFormat,
+            Layout = a.Layout,
+        };
+        await Assert.That(a.ContentHash()).IsNotEqualTo(b.ContentHash());
+    }
+
+    [Test]
+    public async Task name_does_not_participate_in_hash_or_equality()
+    {
+        var a = BuildSample("debug-1");
+        var b = BuildSample("debug-2");
+        await Assert.That(a == b).IsTrue();
+        await Assert.That(a.GetHashCode()).IsEqualTo(b.GetHashCode());
+    }
+}
--- a/src/Paradise.Rendering.Test/RenderCommandEncoderTests.cs
+++ b/src/Paradise.Rendering.Test/RenderCommandEncoderTests.cs
@@ -1,0 +1,64 @@
+using System;
+using System.Buffers;
+
+namespace Paradise.Rendering.Test;
+
+public class RenderCommandEncoderTests
+{
+    [Test]
+    public async Task encoder_round_trip_preserves_command_kinds_and_payloads()
+    {
+        var writer = new ArrayBufferWriter<RenderCommand>(8);
+        var encoder = new RenderCommandEncoder(writer);
+
+        var pipeline = new PipelineHandle(7, 3);
+        var vbuf = new BufferHandle(11, 5);
+        var ibuf = new BufferHandle(13, 5);
+
+        encoder.BeginPass(0);
+        encoder.SetPipeline(pipeline);
+        encoder.SetVertexBuffer(slot: 1, vbuf, offset: 32, size: 256);
+        encoder.SetIndexBuffer(ibuf, IndexFormat.Uint32, offset: 0, size: 64);
+        encoder.SetBindGroup(2);
+        encoder.Draw(new DrawCommand(VertexCount: 3, InstanceCount: 1, FirstVertex: 0, FirstInstance: 0));
+        encoder.DrawIndexed(new DrawIndexedCommand(IndexCount: 6, InstanceCount: 1, FirstIndex: 0, BaseVertex: 0, FirstInstance: 0));
+        encoder.EndPass();
+
+        var memory = writer.WrittenMemory;
+        await Assert.That(memory.Length).IsEqualTo(8);
+
+        var commands = memory.ToArray();
+        await Assert.That(commands[0].Kind).IsEqualTo(RenderCommandKind.BeginPass);
+        await Assert.That(commands[0].BeginPass.PassIndex).IsEqualTo(0);
+
+        await Assert.That(commands[1].Kind).IsEqualTo(RenderCommandKind.SetPipeline);
+        await Assert.That(commands[1].SetPipeline.Pipeline).IsEqualTo(pipeline);
+
+        await Assert.That(commands[2].Kind).IsEqualTo(RenderCommandKind.SetVertexBuffer);
+        await Assert.That(commands[2].SetVertexBuffer.Slot).IsEqualTo(1u);
+        await Assert.That(commands[2].SetVertexBuffer.Buffer).IsEqualTo(vbuf);
+        await Assert.That(commands[2].SetVertexBuffer.Offset).IsEqualTo(32ul);
+        await Assert.That(commands[2].SetVertexBuffer.Size).IsEqualTo(256ul);
+
+        await Assert.That(commands[3].Kind).IsEqualTo(RenderCommandKind.SetIndexBuffer);
+        await Assert.That(commands[3].SetIndexBuffer.Buffer).IsEqualTo(ibuf);
+        await Assert.That(commands[3].SetIndexBuffer.Format).IsEqualTo(IndexFormat.Uint32);
+
+        await Assert.That(commands[4].Kind).IsEqualTo(RenderCommandKind.SetBindGroup);
+        await Assert.That(commands[4].SetBindGroup.GroupIndex).IsEqualTo(2u);
+
+        await Assert.That(commands[5].Kind).IsEqualTo(RenderCommandKind.Draw);
+        await Assert.That(commands[5].Draw.VertexCount).IsEqualTo(3u);
+
+        await Assert.That(commands[6].Kind).IsEqualTo(RenderCommandKind.DrawIndexed);
+        await Assert.That(commands[6].DrawIndexed.IndexCount).IsEqualTo(6u);
+
+        await Assert.That(commands[7].Kind).IsEqualTo(RenderCommandKind.EndPass);
+    }
+
+    [Test]
+    public async Task encoder_throws_on_null_writer()
+    {
+        await Assert.That(() => new RenderCommandEncoder(null!)).Throws<ArgumentNullException>();
+    }
+}
--- a/src/Paradise.Rendering.Test/RenderCommandLayoutTests.cs
+++ b/src/Paradise.Rendering.Test/RenderCommandLayoutTests.cs
@@ -1,0 +1,31 @@
+using System.Runtime.CompilerServices;
+
+namespace Paradise.Rendering.Test;
+
+/// <summary>Locks the actual size of <see cref="RenderCommand"/> against its declared
+/// <c>StructLayoutAttribute.Size</c>. The CLR silently extends an explicit-layout struct when
+/// declared <c>Size</c> is smaller than the field extents — declaring 40 while
+/// <see cref="SetVertexBufferPayload"/> at <c>FieldOffset(8)</c> needs 48 bytes (4 slot + 4 pad
+/// + 16 BufferHandle + 8 offset + 8 size) silently rounds up and produces no diagnostic. This
+/// test catches the mismatch so future payload additions surface size drift immediately rather
+/// than degrading the declared cap into a polite suggestion.</summary>
+public class RenderCommandLayoutTests
+{
+    [Test]
+    public async Task render_command_size_matches_declared_layout()
+    {
+        await Assert.That(Unsafe.SizeOf<RenderCommand>()).IsEqualTo(48);
+    }
+
+    [Test]
+    public async Task largest_payload_set_vertex_buffer_fits_within_struct()
+    {
+        // SetVertexBufferPayload = (uint Slot, BufferHandle Buffer, ulong Offset, ulong Size)
+        //   = 4 (Slot) + 4 (alignment pad) + 16 (BufferHandle is StructLayout.Size=16)
+        //     + 8 (Offset) + 8 (Size) = 40 bytes
+        // Placed at FieldOffset(8), the struct needs at least 48 bytes total.
+        var payloadSize = Unsafe.SizeOf<SetVertexBufferPayload>();
+        await Assert.That(payloadSize).IsEqualTo(40);
+        await Assert.That(Unsafe.SizeOf<RenderCommand>()).IsGreaterThanOrEqualTo(payloadSize + 8);
+    }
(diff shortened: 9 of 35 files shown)


quabug:
## Iteration 2 — fixes pushed (commit `47ea578`)

### CI race blocker
- `tools/slang/SlangBootstrap.cs` promoted to a real csproj. Previous `dotnet run --file` shared the dotnet runfile cache (`~/.local/share/dotnet/runfile/<scriptHash>/obj/...AssemblyInfoInputs.cache`) across consumer projects → MSB3491 in CI on parallel project builds. Scoped csproj isolates artifacts under `tools/slang/bin/obj` and MSBuild's per-project locking serializes naturally. `_BuildSlangBootstrap` dispatches via `<MSBuild>` with `RemoveProperties` for inherited globals (`TreatWarningsAsErrors`, `EnableDefaultCompileItems`, framework/RID) so the bootstrap project sees a clean property bag.

### Major findings (all 3 reviewers)
- **Shader leak / cache miss in `CreatePipeline(ShaderProgramDesc, TextureFormat)`** (R1 minor / R2 major / R3 major): `WebGpuDevice.CreateShaderModule` now dedupes by `(Wgsl, EntryPoint, Stage)` via a content-keyed cache. Two `ShaderProgramDesc` loads of the same logical program return the same `ShaderHandle` → `PipelineDesc.ContentHash` is stable across calls → pipeline cache hits → shader slots stop accumulating. `DestroyShader` drops the cache entry so the next `CreateShaderModule` on the same tuple compiles a fresh module.
- **Depth/stencil silently dropped** (R2 major): `WebGpuDevice.CreatePipeline` throws `NotSupportedException` for non-null `DepthStencilFormat`; `WebGpuRenderer.BeginPass` does the same for non-null `RenderPassDesc.Depth`. Field surface is reserved for M2 (#43) but the M1 backend has no plumbing — explicit rejection beats half-supported behavior.

### Minor findings
- **PipelineCache collision orphans the displaced entry** (R1 minor): `GetOrCreate` accepts an optional `onEvict` callback fired when a structurally-different desc collides on the same hash bucket. Wired in `WebGpuRenderer` to schedule the displaced pipeline on the deferred destruction queue.
- **Nested `BeginPass` silently overwrites active pass** (R3 minor): `ExecuteStream` throws when `BeginPass` is encountered while `activePass` is non-null, mirroring the post-loop missing-EndPass guard.
- **Leaked render-pass encoder on mid-pass exception** (R3 minor): `ExecuteStream` wraps the dispatch loop in `try/finally` so a stale-handle / `NotSupportedException` mid-pass calls `activePass.End()` before the encoder is finished.
- **Per-frame allocation in `TriangleScene`** (R3 suggestion): `ArrayBufferWriter<RenderCommand>` hoisted to a field; `ResetWrittenCount()` per frame. Sample's render loop is now allocation-free at steady state.
- **Unused `using` alias** (R3 suggestion): dropped `WgWebGpuManagedSpan` from `WebGpuDevice.cs`.

### Deferred (with rationale)
- **`<SlangVersion>` regex fragility** (R3 risk): the manifest schema today has a single top-level `"version"` key; a pinned tool version is the simplest invariant. If a future schema adds per-RID metadata, the rename to `"toolVersion"` (or a JSON peek) lands with that schema bump — not in this PR.
- **`DefaultFramesInFlight = 2` vs Dawn's actual frame depth** (R3 risk): dependent on Dawn backend selection; needs runtime probing rather than a hardcode change. M2 will revisit alongside the device-loss split (#54).
- **Recursive shader-glob basename collision** (R2 minor): documented in `Slang.targets` as a follow-up — M1/M2 only author flat shader directories. The fix needs `%(RecursiveDir)` propagation through both `slangc` output paths and `LogicalName`; deferring to whichever issue first introduces nested authoring.
- **Loader-output `PipelineLayoutDesc` reference inequality breaks pipeline cache for repeated `CreatePipeline(ShaderProgramDesc, …)` calls across ShaderProgramLoader.Load invocations** (R1 suggestion): the shader-module cache fix neutralizes the practical impact for M1 (single `ShaderProgramDesc` load → single pipeline). M2 will revisit when dynamic pipeline rebuilds become common (could content-hash the layout instead of reference-comparing).

### Verification
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` — 0 errors
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false` — 1972/1972 pass (1 new test: `two_loads_produce_modules_with_identical_wgsl_and_distinct_records` locks the loader property the device dedupe relies on)
- `dotnet publish src/Paradise.Rendering.Sample -c Release` — AOT clean
- Headless AOT binary still renders the triangle and exits 0
- Local parallel build no longer races (`tools/slang/bin/obj` cleared first; both Sample + WebGPU.Test build in one go without MSB3491)

quabug:
@opencara review

quabug:
## Iteration 2 — additional commits since the prior trigger

Three commits landed after `47ea578` was triggered for review at 10:46Z; surfacing them so the iteration-2 verdict reflects the full final state of the PR.

### `0018a72` — CI race fix #2 (file lock)
The `47ea578` change to a csproj-based `SlangBootstrap` removed the `dotnet run --file` runfile cache race, but exposed a second race: two MSBuild `Exec` invocations of the bootstrap binary (Sample + WebGPU.Test in parallel) raced on the destination cache dir, both writing `slang-archive.tar.gz` simultaneously → IOException ("being used by another process"). Fixed by adding a cross-process `FileShare.None` lock on `<cacheRoot>/.bootstrap.lock` at the top of the bootstrap. Second invocation blocks until the first releases, then short-circuits via the existing `.installed` marker. CI run on `0018a72` succeeded.

### `f35295f` — explicit `LoadOp`/`StoreOp` switch mapping
Addresses R1's "Question" finding about the binary `src.Load == LoadOp.Load ? Load : Clear` pattern silently falling through to `Clear` for any future enum value. Replaced with `switch` expressions that throw `NotSupportedException` on unrecognized values — behavior-preserving for the current 2-value enums but turns a future enum addition into a build-break instead of a silent miscompile.

### `f152f16` — Snapshot B raw-reflection embed + insert-only `PipelineCache`
- **Snapshot B for #45**: after slangc emits `triangle.reflection.json`, copy it to `triangle.raw-reflection.json` and embed both as `<EmbeddedResource>` with distinct LogicalNames (`Shaders.<name>.reflection.json` and `Shaders.<name>.raw-reflection.json`). Today the runtime loader transforms the same raw JSON at load time so the two embedded resources are byte-identical; when the transform moves to build-time, `reflection.json` diverges and #45's two-snapshot golden test catches the divergence. Tests for Snapshot B stay in #45 — this PR ships only the embed plumbing. Verified: `Paradise.Rendering.Sample.dll` manifest lists both names.
- **PipelineCache → insert-only**: replaced the `onEvict` callback (added in iteration 2 to handle hash collisions) with insert-only semantics — a 32-bit hash collision between two structurally-different `PipelineDesc` instances throws `InvalidOperationException` instead of silently overwriting or invoking a "caller must destroy displaced" callback. Removes a behavioral contract from the public surface that downstream code (#43+) would otherwise have had to remember; callers who genuinely intend to replace call `Forget()` first.

### Verification
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` — clean
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false` — 1972/1972 pass
- `dotnet publish src/Paradise.Rendering.Sample -c Release` — AOT clean
- Headless AOT binary renders triangle, exits 0
- Manifest resources: `Shaders.triangle.wgsl`, `Shaders.triangle.reflection.json`, `Shaders.triangle.raw-reflection.json`

quabug:
@opencara review

quabug:
## Iteration 3 — fixes pushed (commit `4538636`)

Addresses every actionable finding from the iteration-2.5 verdict against `f152f16`. CI on `4538636` is green. Local: 1976/1976 tests pass, AOT publish clean, headless triangle still renders.

### Major findings — fixed

#### `WebGpuRenderer.BeginPass` missing Depth guard
Added symmetric `NotSupportedException` for non-null `RenderPassDesc.Depth`, mirroring the existing guard in `WebGpuDevice.CreatePipeline` for non-null `PipelineDesc.DepthStencilFormat`. The iteration-2 commit message claimed both halves landed; only the pipeline-side half did. Regression test: `HandleDistinctnessTests.begin_pass_rejects_non_null_depth_attachment`.

#### `DestroyShader` cache-eviction race
Restructured so the dedupe-cache eviction happens SYNCHRONOUSLY at schedule time, while the slot-table `Remove` stays deferred. `WebGpuDevice` exposes a new `EvictShaderFromCache` method (cache-only); `WebGpuRenderer.DestroyShader` calls it before `_destructionQueue.Schedule(...)`. Closes the use-after-free window where a `CreateShader` with the same `(Wgsl, EntryPoint, Stage)` tuple between schedule and the deferred callback would resolve to the dying handle. Regression test: `HandleDistinctnessTests.destroy_shader_then_recreate_with_same_content_returns_distinct_handle`.

#### `PipelineCache` redesigned per opus's recommendation: cache below the public handle layer
This is a deliberate design refinement, not a point fix. Iteration 2 had the cache returning the same `PipelineHandle` to two callers with structurally-equal descs — first caller's `DestroyPipeline` invalidated the second caller's still-live handle (contract violation; every other resource type — `BufferHandle`, `TextureHandle`, `ShaderHandle` — preserves "destroy invalidates only that handle"). The restructure:
- `PipelineCache` now caches native `WebGpuSharp.RenderPipeline` instances keyed on `PipelineDesc.ContentHash`
- `WebGpuDevice.CreatePipeline` split into `BuildNativePipeline` (creates a native pipeline, no slot allocation) + `RegisterPipeline` (slot allocation only, takes an existing native pipeline)
- `WebGpuRenderer.CreatePipeline` (a) gets-or-creates the cached native via the cache, then (b) mints a fresh `PipelineHandle` via `RegisterPipeline`
- Two callers with structurally-equal descs share the GPU compile but receive distinct `PipelineHandle`s; `DestroyPipeline(h1)` does not invalidate `h2`
- Cache lives for renderer lifetime (no refcount, no LRU eviction); revisit when M2/M3 introduces dynamic pipeline rebuilds

Regression test: `HandleDistinctnessTests.two_create_pipeline_calls_return_distinct_handles_with_shared_native_cache` — verifies distinct handles AND that destroying one doesn't invalidate the other (the surviving handle is exercised through a real `Submit`).

### Minor — fixed

#### `RenderCommand` declared `Size = 40` was smaller than the largest payload
Corrected to `Size = 48`. `SetVertexBufferPayload` at `FieldOffset(8)` is 40 bytes (4 Slot + 4 align + 16 BufferHandle + 8 Offset + 8 Size) so the struct needs at least 48 bytes; CLR was silently extending. Fixed declaration, corrected the doc comment to reference the 40-byte payload, added `RenderCommandLayoutTests` (`Unsafe.SizeOf<RenderCommand>() == 48` plus a sanity check that the SetVertexBuffer payload actually fits).

### Suggestion — applied

#### Drop dead `ReferenceEquals(vsModule, fsModule)` branch in `WebGpuRenderer.CreatePipeline(ShaderProgramDesc, ...)`
The loader always allocates distinct `ShaderModuleDesc` records per entry point so the conditional was unreachable. Removed; comment explains the dedupe is in the device's content-keyed cache.

### Risks deferred (per prior coordination)

- `PipelineDesc.Layout` reference-equality — M2 follow-up, file as a separate issue when the loader's Layout instances start being shared.
- WebGPUSharp 0.5.2 `CreateRenderPipelineSync` vertex-attribute-span lifetime — requires a WebGPUSharp source check; if confirmed an issue, file upstream against WebGPUSharp.

### Reviewer-flagged items from iteration-2.5

- `qwen3.5-plus/qwen` — "critical memory corruption" framing is wrong per opus's analysis; the underlying observation is the minor doc/declaration fix above.
- `glm-5/qwen` — quoted pre-iteration-2 LoadOp/StoreOp pattern already fixed in `f35295f`.

### Test count

1972 → 1976. Net delta from iteration-2.5: +4 (3 `HandleDistinctnessTests` + 2 `RenderCommandLayoutTests` − 1 `PipelineCacheTests` removed since the cache type now parameterizes on `WebGpuSharp.RenderPipeline` which has no public constructor for unit-test stubbing; the cache contract is now exercised by the integration-level distinctness tests, which is a stronger guarantee).

quabug:
@opencara review

quabug:
## Iteration 4 — response to iteration-3 verdict (4538636)

Pushed 705237d addressing all three defects from the iteration-3 OpenCara verdict.

### Major 1 — Destroy* synchronous slot eviction
Previously `DestroyShader` / `DestroyBuffer` / `DestroyPipeline` all deferred slot-table removal on the `DeferredDestructionQueue`, so destroyed handles stayed resolvable for `DefaultFramesInFlight` frames. Split the operation into two halves:
- New internal `SlotTable.Detach(index, generation, out value)` atomically invalidates the slot (nulls value, bumps generation, pushes the index onto the free list) and hands the native object back to the caller.
- `WebGpuRenderer.Destroy*()` calls `Detach*()` synchronously, then schedules ONLY the native GPU release on the deferred-destruction queue (via a closure that retains the native reference).
- Pipelines are slot-invalidated inline (no native teardown to defer — the `WgRenderPipeline` is cache-owned).

After `Destroy*()` returns, `Resolve*()` throws `StaleHandleException` immediately. Two new regression tests (`destroy_buffer_invalidates_handle_synchronously`, `destroy_pipeline_invalidates_handle_synchronously`) drive the full Submit path with a destroyed handle and assert the exception.

### Major 2 — MappedAtCreation removed
Per the PM summary (\"err toward removal if the map/unmap API would widen scope beyond M1\"), dropped `MappedAtCreation` from the public `BufferDesc` surface. `CreateBufferWithData` was already initialising via `Queue.WriteBuffer`, not a mapped range, so the flag was unreachable. `WebGpuDevice.CreateBuffer` no longer passes it to `WgBufferDescriptor` and the `TriangleScene` sample has been updated. A proper map/unmap API lands in a later milestone together with the flag.

### Minor 3 — Structural equality on PipelineLayoutDesc
`PipelineDesc.Equals` / `ContentHash` now walk `PipelineLayoutDesc` structurally (bind-group entries + push-constant ranges) instead of using `ReferenceEquals`. `ShaderProgramLoader.BuildProgramDesc` mints a fresh `PipelineLayoutDesc` per load, so reference identity defeated the pipeline cache — two loads of the same shader retained duplicate native pipelines. Added `layout_is_compared_structurally_not_by_reference` and `layout_structural_inequality_is_detected` tests.

### Build / test status
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` → 0 errors.
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false` → **1980 / 1980 passing** (+4 net new tests over iteration 3's 1976).

quabug:
@opencara review

quabug:
## Iteration 5 — response to iter-4 verdict (705237d → e577f52)

Pushed e577f52 addressing all four iter-4 items.

### Major A — Shader public-handle split
Matches the iter-4 pipeline fix. Before: `CreateShader(ShaderModuleDesc)` interned the public `ShaderHandle` by content key, so two callers got the same handle and either caller's `DestroyShader` invalidated the other's live handle. Now:
- `WebGpuDevice._shaderModuleCache` stores **native `WgShaderModule`** values keyed by `(Wgsl, EntryPoint, Stage)` — the content dedupe lives BELOW the public handle layer.
- Every `CreateShaderModule(desc)` call allocates a fresh `ShaderHandle` via `Shaders.Add(native)`, pointing at the shared cached native.
- `DestroyShader` is pure slot invalidation: it detaches just that one handle and never touches the content cache, because another slot may still reference the same native.
- Deleted the obsolete `EvictShaderFromCache` + reverse-lookup scan (now unreachable).
- Raw `CreateShader(string wgsl, string label)` still mints a fresh native per call (uncached path for tests and non-Slang consumers).

Covered by `two_create_shader_module_calls_return_distinct_handles_with_shared_native_cache`: creates two shader handles for the same `ShaderModuleDesc`, destroys one, builds a pipeline that references the other — verifies `h2` is still live via a successful `CreatePipeline`.

### Major B — Reject non-empty `PipelineDesc.Layout`
Per PM Option A (M1 scope), added a `NotSupportedException` guard in `WebGpuDevice.BuildNativePipeline`:

```csharp
if (desc.Layout is { } l && (l.Groups.Length > 0 || l.PushConstants.Length > 0))
    throw new NotSupportedException(
        "PipelineDesc.Layout with non-empty bind groups or push constants is reserved for M2 (#43). " +
        "The M1 backend uses Dawn's auto layout from the shader module; explicit layouts are not " +
        "plumbed through yet. Pass a layout with empty Groups and PushConstants (or null) until M2.");
```

Empty layouts (`Groups=[]`, `PushConstants=[]` — what `ShaderProgramLoader.BuildProgramDesc` produces) are still accepted. The property and structural equality/hash stay intact so M2 can plumb explicit layouts without an API-shape change.

Covered by `create_pipeline_rejects_non_empty_layout`.

### Minor — `SlotTable.Count` invariant (rebutted with proof tests)
I believe this finding is mistaken. `SlotTable.Count` is not a stored field — it's `_slots.Count - _free.Count`. Both `Remove` and `Detach` push onto `_free`, so the derived count decrements correctly on either path. The finding assumed `Remove` did `_count--`; it doesn't — check the source. No code change in `SlotTable.cs`.

To pin the invariant empirically (so any future refactor to a stored `_count` field can't reintroduce the imagined drift) I added three `SlotTableTests` cases:

- `detach_returns_value_and_invalidates_old_handle` — the happy-path detach contract.
- `detach_then_count_reports_zero` — directly rebuts the finding: `Add` then `Detach` leaves `Count == 0`.
- `detach_with_stale_handle_returns_false` — double-detach / wrong-generation / out-of-range.

These tests are also device-free, which addresses the secondary iter-4 risk (the GPU-only `HandleDistinctnessTests` are skipped on CI without an adapter; the core synchronous-detach contract is now pinned in a device-free test that always runs).

### Suggestion — `GC.KeepAlive` cleanup
Replaced `_destructionQueue.Schedule(() => { GC.KeepAlive(native); });` with `_destructionQueue.Schedule(() => { _ = native; });` and updated the comment. `GC.KeepAlive` only prevents elision of stack-allocated locals in its enclosing method; applied to a heap-allocated closure field it has no runtime effect. The closure capture alone roots the native reference until the deferred frame fires — the explicit `GC.KeepAlive` was misleading.

### Note on iter-5 cache interaction
Because `CreatePipeline(ShaderProgramDesc, TextureFormat)` now mints fresh `ShaderHandle`s per call (shader cache is below the public handle layer), two invocations of the high-level helper produce distinct content-hashes → the `PipelineCache` misses between them. The low-level `CreatePipeline(in PipelineDesc)` still hits the cache whenever a caller reuses shader handles. The `PipelineCache` itself is unchanged; the change moves the caching burden one layer down. Trade-off acknowledged — documented in the test's doc-comment — and preferred over the public-handle-sharing bug.

### Build / test status
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` → **0 errors**.
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false` → **1985 / 1985 passing** (+5 net new).

quabug:
@opencara review

quabug:
## Iteration 6 — response to iter-5 verdict (e577f52 → 6718838)

Pushed 6718838 addressing both iter-5 minor findings plus a regression test.

### Minor 1 — `CreatePipeline(ShaderProgramDesc, TextureFormat)` shader-slot leak
Applied the 2-line destroy-after-build pattern per OpenCara's recommendation:

```csharp
var pipeline = CreatePipeline(in pipelineDesc);
DestroyShader(vsHandle);
DestroyShader(fsHandle);
return pipeline;
```

The locally-minted `vsHandle`/`fsHandle` are now released as soon as the native `WgRenderPipeline` has been built. The native `WgShaderModule` stays alive through (a) `_shaderModuleCache` content-keyed retention and (b) the native pipeline's own layout reference — so post-creation destruction affects only `_device.Shaders` slot-table metadata, never the native resource.

The deeper pipeline-cache miss (content hash folds per-call handle indices) is acknowledged as an M2 follow-up per PM guidance ("M2 better fix is content-derived PipelineCache keys — out-of-scope for iter-6").

### Minor 2 — Stale comment
Rewrote the comment block at `WebGpuRenderer.cs:236` so it matches current behavior. It now describes the iter-5 content-cache-below-handle-layer semantics and explains why the two locally-minted handles must be destroyed before the helper returns.

### New regression coverage
- Added `internal int ShaderSlotCountForTest => _device.Shaders.Count;` on `WebGpuRenderer` so the slot-table non-growth property can be asserted without reflection or a GPU-specific harness.
- `HandleDistinctnessTests.create_pipeline_from_program_does_not_grow_shader_slot_table`: warms the pipeline cache with one call, records the baseline slot count, then runs 8 more `CreatePipeline(program, fmt)` invocations and asserts the count is unchanged. Any regression (forgotten `DestroyShader`, inline refactor that drops the destroy pair) fails hard.

### Non-blocking items intentionally NOT addressed
Per PM guidance and the OpenCara review's "take if trivial, else defer" tagging:

- `DestroyShader` deferred closure inertia — the content cache retaining natives for renderer lifetime makes the `_destructionQueue.Schedule` call inert. Deferred until M2 introduces cache eviction so the closure becomes load-bearing. Leaving in place; no-op cost is negligible.
- `CreateBufferWithData` silent size enlargement — M2 API-hardening item.
- `WebGpuManagedSpan` vertex-attribute lifetime — needs WebGPUSharp source confirmation; out-of-scope.
- `Slang.targets` basename flattening — already documented in-file as an M2+ follow-up.

### Build / test status
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` → **0 errors**.
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false` → **1986 / 1986 passing** (+1 net new).

quabug:
@opencara review

quabug:
## Iteration 7 — response to iter-6 verdict (6718838 → afa8dc3)

Pushed afa8dc3 addressing the two majors (gpt-5.4/codex) and the minor (claude-sonnet-4-6) from iter-6.

### Major A — Loader rejects non-varying shader bindings
`ShaderProgramLoader.BuildProgramDesc` used to silently emit an empty `PipelineLayoutDesc` regardless of whether Slang reflected uniforms / storage buffers / samplers / textures / push constants. Fixed by adding `RejectNonVaryingBindings`: scans each entry point's parameter list and throws `NotSupportedException` with an M2 deferral message if any `Binding.Kind` isn't `varyingInput` / `varyingOutput`. Empty layouts (triangle's shape) pass unchanged — `triangle_reflection_round_trips` and the existing loader tests still succeed.

Symmetric with the `PipelineDesc.Layout` guard already in `WebGpuDevice.BuildNativePipeline` (iter-5) — two layers now reject unsupported binding data at load-time and at pipeline-build-time.

### Major B — `CreatePipeline(ShaderProgramDesc, TextureFormat)` topology parameters
Added defaulted parameters so non-triangle-list callers can override:

```csharp
public PipelineHandle CreatePipeline(
    in ShaderProgramDesc program,
    TextureFormat colorFormat,
    PrimitiveTopology topology = PrimitiveTopology.TriangleList,
    IndexFormat stripIndexFormat = IndexFormat.Uint16)
```

Defaults match the prior hardcoded behavior, so the sample (`TriangleScene.cs:38`) and every other existing call site stays source-compatible. Point/line/strip callers now pass `topology:` explicitly instead of getting wrong primitive assembly silently.

### Minor — `FormatConversions.FromWgpu(TextureFormat)` throws on unknown
Changed the fallthrough arm from `_ => TextureFormat.Undefined` to `_ => throw new NotSupportedException($"WebGPU texture format '{f}' has no Paradise.Rendering mapping.")`. Symmetric with every sibling `ToWgpu` helper in the same file. HDR swapchain formats or future WebGPUSharp values now surface at the conversion site instead of producing an `Undefined` pipeline color format that Dawn later rejects opaquely.

### New tests (+6)
- `ShaderProgramLoaderTests.build_program_desc_rejects_non_varying_bindings` — Major A
- `ShaderProgramLoaderTests.build_program_desc_accepts_varying_input_and_output` — Major A positive
- `HandleDistinctnessTests.create_pipeline_from_program_respects_custom_topology` — Major B (GPU-gated, exercises `PrimitiveTopology.PointList` through the helper)
- `FormatConversionsTests.from_wgpu_throws_not_supported_for_unmapped_value` — Minor
- `FormatConversionsTests.from_wgpu_round_trips_the_mapped_bgra8unorm_value` — Minor positive
- `FormatConversionsTests.from_wgpu_maps_undefined_symmetrically` — Minor positive

### Non-blocking items intentionally NOT addressed
Per PM scope guidance and OpenCara's iter-6 "take if trivial, else defer" tagging:
- `DestroyShader` deferred closure inertia — M2 cache-eviction dependency.
- `CreateBufferWithData` silent size enlargement — M2 API hardening.
- `PipelineCache` content-derived keys (so repeated `CreatePipeline(ShaderProgramDesc, fmt)` calls actually hit) — M2 follow-up.
- `WebGpuManagedSpan` vertex-attribute lifetime — needs WebGPUSharp source confirmation; out-of-scope.
- `Slang.targets` basename flattening — already documented in-file as an M2+ follow-up.
- 32-bit `ContentHash()` collision ceiling — M3+ concern (birthday threshold ~65k pipelines).

### Build / test status
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` → **0 errors**.
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false` → **1992 / 1992 passing** (+6 net new).

quabug:
@opencara review

quabug:
## Iteration 8 — response to iter-7 verdict (afa8dc3 → df244bf)

Pushed df244bf addressing the three polish items from iter-7.

### Minor — `ExecuteStream` double-`End()` on `EndPass` throw
Rewrote the `EndPass` case so `activePass` is nulled **before** `End()` is invoked:

```csharp
case RenderCommandKind.EndPass:
{
    var passToEnd = activePass;
    activePass = null;
    passToEnd?.End();
    break;
}
```

The finally-block safety net (`activePass?.End()` at function exit) is now idempotent — if Dawn throws at pass end, `activePass` is already null and we don't double-end the native encoder.

### Minor — Shader-handle leak on `CreatePipeline(ShaderProgramDesc)` exception path
Wrapped the inner pipeline call in try/finally so `DestroyShader(vsHandle)` + `DestroyShader(fsHandle)` always run, even when `BuildNativePipeline`'s `DepthStencilFormat` / `Layout` guards throw. Same defect class iter-6 closed for the happy path; exception path now also covered.

### Suggestion — `CreateBufferWithData` int overflow
`(ulong)(data.Length * Unsafe.SizeOf<T>())` computed the multiplication in `int` first, silently wrapping at 2^31. Widened both operands before multiplying:
```csharp
var byteSize = (ulong)data.Length * (ulong)Unsafe.SizeOf<T>();
```

Not reachable from M1's triangle path (15 floats) but a sharp edge for future staging spans.

### New regression test
- `HandleDistinctnessTests.create_pipeline_from_program_releases_shader_handles_on_exception` — constructs a `ShaderProgramDesc` with a non-empty `PipelineLayoutDesc` so the inner `BuildNativePipeline` guard throws, then asserts `ShaderSlotCountForTest` didn't grow across four rejected calls.

### Acknowledged M2 deferrals (unchanged stance)
Per OpenCara explicitly marking these as acknowledged/deferred across iter-5/6/7 and PM's scope guidance:
- `PipelineCache` growth per `CreatePipeline(ShaderProgramDesc)` call — M2 content-derived cache keys.
- `PipelineDesc.Equals` / `ContentHash` including `StripIndexFormat` + null/empty `Layout` — M2 normalization.
- `DestroyShader` deferred closure inertia — M2 cache eviction.
- `WebGpuManagedSpan` vertex-attribute lifetime — needs WebGPUSharp source check.
- `Slang.targets` basename flattening — documented M2 item.

### Build / test status
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` → **0 errors**.
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false` → **1993 / 1993 passing** (+1 net new).

quabug:
@opencara review

quabug:
## Iteration 9 — response to iter-8 verdict (df244bf → bad7fdf)

Pushed bad7fdf closing both iter-8 items.

### Minor — `vsHandle` leak if the second `CreateShaderModule` throws
Widened the try/finally to cover both module allocations plus the `PipelineDesc` construction plus the inner `CreatePipeline`. Default-init both handles and guard each destroy with `IsValid`:

```csharp
ShaderHandle vsHandle = default;
ShaderHandle fsHandle = default;
try
{
    vsHandle = _device.CreateShaderModule(vsModule);
    fsHandle = _device.CreateShaderModule(fsModule);
    var pipelineDesc = new PipelineDesc { ... };
    return CreatePipeline(in pipelineDesc);
}
finally
{
    if (vsHandle.IsValid) DestroyShader(vsHandle);
    if (fsHandle.IsValid) DestroyShader(fsHandle);
}
```

`default(ShaderHandle).Generation == 0` and `IsValid = Generation != 0`, so unallocated handles are safely skipped.

### Suggestion (cosmetic) — `ExecuteStream` try body reindent
Reindented the for-loop body one level deeper so it reads as "inside the try" rather than "adjacent to the try brace." Behavior unchanged; just Allman alignment consistent with the rest of the file.

### New regression test
- `HandleDistinctnessTests.create_pipeline_from_program_releases_vs_handle_when_fs_create_throws` — tampered `ShaderProgramDesc` pairs real VS WGSL with deliberately invalid FS WGSL. Dawn rejects the FS module, `WebGpuDevice.CreateShaderModule` throws `InvalidOperationException("ShaderModule creation returned null.")`, the iter-9 finally hits with `vsHandle` allocated + `fsHandle` still default. Asserts `ShaderSlotCountForTest` holds across 4 repeated rejected calls — any regression to "allocations outside try" reappears as baseline drift.

### Acknowledged M2 deferrals (unchanged)
- `PipelineCache` growth per `CreatePipeline(ShaderProgramDesc)` call.
- `PipelineDesc.Equals`/`ContentHash` including `StripIndexFormat` + null/empty `Layout`.
- `DestroyShader` deferred closure inertness for the Slang path.
- `WebGpuManagedSpan` vertex-attribute lifetime — needs WebGPUSharp source check.
- `Slang.targets` basename flattening — documented M2 item.

### Build / test status
- `dotnet build ParadiseEngine.slnx -p:TreatWarningsAsErrors=true` → **0 errors**.
- `dotnet test --solution ParadiseEngine.slnx -p:PublishAot=false` → **1994 / 1994 passing** (+1 net new).

quabug:
@opencara review

